PR #21, opened 2021-03-11
Add CI (github actions); add third_party/python/markupsafe.

tcal-x:
CI runs two jobs in parallel -- check-docs and check-sw.    "check-sw" compiles the software for `proj/mnv2_first`, requiring that it goes through all of the CFU-Playground setup and also manually installs a good RISC-V toolchain.


tcal-x:
check-docs is redundant since a PR also triggers a doc build on readthedocs.   I will probably remove it.

scripts/setup-ci
```diff
@@ -0,0 +1,47 @@
...
+fi
+
+if [ ${#missing[@]} -gt 0 ]; then
+    echo "Missing packages: ${missing[@]}" >&2
+    sudo apt install -y "${missing[@]}"
+fi
+
+# Warn about different versions
+
+if ! riscv64-unknown-elf-gcc --version | grep ' 8.3.' >/dev/null; then
```
alanvgreen, line 34: The toolchain I'm using has this version:

riscv64-unknown-elf-gcc (SiFive GCC 10.1.0-2020.08.2) 10.1.0

Maybe we need to update scripts/setup too.

.github/workflows/sample.yml
```diff
@@ -0,0 +1,26 @@
+name: sample
```
alanvgreen, line 1: A more descriptive name? (same for filename)

scripts/setup-ci
```diff
@@ -0,0 +1,47 @@
+#!/bin/bash
```
alanvgreen, line 1: Can this file be integrated with scripts/setup?

It would be a pity if they fell out of sync.

alanvgreen (approved):
Am happy for this to go in as-is, and these comments can be addresed later.

scripts/setup-ci
tcal-x: Oh, so 8.3 is good, 9.x is bad, but 10.1 is good again?   I'll have to check that out.

.github/workflows/sample.yml
tcal-x: Yep, done.

PR #22, opened 2021-03-12
Swcpu1

alanvgreen:
Move filter values into EBRAM


proj/mnv2_first/gateware/mnv2_cfu.py
```diff
@@ -13,14 +13,17 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+from types import MemberDescriptorType
+from nmigen.hdl.ast import _MappedKeyCollection
```
tcal-x, line 17: Where is _MappedKeyCollection used?

tcal-x:
@alanvgreen , I see the mnv2 "golden" test fail with this.   Is that expected?

alanvgreen:
> @alanvgreen , I see the mnv2 "golden" test fail with this. Is that expected?

It was working for me. Will investigate!

tcal-x:
To make sure I didn't have something messed up, I checked out 'avg12' and ran that --- with that, the golden tests passed.

proj/mnv2_first/gateware/mnv2_cfu.py
alanvgreen: Not sure. Will fix in the next PR.

alanvgreen:
I think this is good to merge now.

tcal-x:
I found a very long path in the timing report ending at `VexRiscv/Cfu/fn0/fvg/count[2047]_i_2_n_0`.    The length of the chain seemed to be suspicious.   In the nMigen, we found a line `self.limit = Signal(self.num_memories * depth)` that may be incorrect.  Perhaps it should be `self.limit = Signal(range(self.num_memories * depth))`.

tcal-x:
After I made that fix, everything worked on the board for me, so that does seem to have been the issue (a code mistake creating a 2048b counter).

proj/mnv2_first/gateware/store.py
```diff
@@ -0,0 +1,186 @@
...
+    """
+
+    def __init__(self, num_memories, depth):
+        super().__init__()
+        assert 0 == ((num_memories - 1) &
+                     num_memories), "Num memories (f{num_memories}) not a power of two"
+        self.num_memories = num_memories
+        self.r_addr = Signal(range(depth))
+        self.r_bank = Signal(range(num_memories))
+        self.limit = Signal(self.num_memories * depth)
```
tcal-x, line 161: Here is the issue.  `self.limit` is used below to determine with width of `count`.

PR #23, opened 2021-03-12
Add Verilator simulation (tflm unit test) to CI.

tcal-x:
The interact.expects script is updated to handle either simulation or Arty board.

CI invokes "make PLATFORM=sim unit", which invokes the expect script, which spawns and interacts with "make PLATFORM=sim load", and then kills it.

### DIFF
--- a/.github/workflows/check-doc-sw.yml
+++ b/.github/workflows/check-doc-sw.yml
@@ -24,3 +24,4 @@ jobs:
       - run: which pip3 && which python3 && which pip
       - run: riscv64-unknown-elf-gcc --version
       - run: pwd && source environment && cd proj/mnv2_first && pip3 list && make -j9 software
+      - run: pwd && source environment && cd proj/mnv2_first && pip3 list && make PLATFORM=sim unit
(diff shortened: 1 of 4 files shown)


.github/workflows/check-doc-sw.yml
```diff
@@ -24,3 +24,4 @@ jobs:
       - run: which pip3 && which python3 && which pip
       - run: riscv64-unknown-elf-gcc --version
       - run: pwd && source environment && cd proj/mnv2_first && pip3 list && make -j9 software
+      - run: pwd && source environment && cd proj/mnv2_first && pip3 list && make PLATFORM=sim unit
```
alanvgreen, line 27: Add -j9 here, too. 

Adding PLATFORM=sim wil force all the C++ to be rebuilt and -j9 will help with that.

alanvgreen (approved):
Just one small comment.

.github/workflows/check-doc-sw.yml
tcal-x: Yes, made this change.   Both make's now have -j8.

PR #28, opened 2021-03-15
Maccs

alanvgreen:
Adds a MACC4 instruction that returns a result to the CPU.

### DIFF
--- a/proj/mnv2_first/Makefile
+++ b/proj/mnv2_first/Makefile
@@ -17,7 +17,7 @@
 export DEFINES :=
 
 # Uncomment this line to use software defined CFU functions in software_cfu.cc
-#DEFINES += CFU_SOFTWARE_DEFINED
+DEFINES += CFU_SOFTWARE_DEFINED
 
 # Uncomment this line to skip debug code (large effect on performance)
 DEFINES += NDEBUG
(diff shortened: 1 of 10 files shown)


tcal-x:
Hi @alanvgreen , I'm not sure if this is a new path or an old one, but there's a pretty long timing path that misses timing by over 7ns.   It starts at a register, looks like it goes through a long carry chain (VexRiscv/Cfu/fn0/ppx/pp/$42_carry__2/CO[3] is an example element on the path), and back to the VexRiscv core.  These are some more elements towards the end of the path:
```
VexRiscv/Cfu/fn0/set_activation_min/stored_output[31]_i_11/O
VexRiscv/Cfu/fn0/f7buf/stored_output_reg[31]_0

```

It maybe that this path was there before, in which case it has nothing to do with this PR.   I'll check.

tcal-x:
> Adds a MACC4 instruction that returns a result to the CPU.

Hi @alanvgreen ,
Can you explain a little about the purpose of these?   Is it for debugging?

tcal-x:
@alanvgreen : I see mismatches running menu item "2 c" -- "Functional CFU Tests" --> "Run hw/sw compare tests".   Could the sw versions be out of sync?   Just let me know if I should ignore this.

tcal-x (approved):
LGTM

PR #33, opened 2021-03-20
Macc4

alanvgreen:
This Macc 4 Run 1 instruction moves the inner accumulation loop into the CFU. The Macc4 accumulates over all input channels.

With this change, we are achieving >1 MACC/cycle over  1x1 2D convs.

The next step will be to integrate the post processing so that the instruction produces an output value ready to store.

### DIFF
--- a/proj/mnv2_first/Makefile
+++ b/proj/mnv2_first/Makefile
@@ -17,7 +17,7 @@
 export DEFINES :=
 
 # Uncomment this line to use software defined CFU functions in software_cfu.cc
-#DEFINES += CFU_SOFTWARE_DEFINED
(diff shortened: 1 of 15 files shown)


tcal-x:
@alanvgreen , I assume this passed timing in your build?   It is failing for me just by 0.15ns using Vivado 2019.1.   I guess I should upgrade.

alanvgreen:
@tcal-x Yes, it passes timing for me. WNS=0.014 so it's pretty close. 

It might be time to start cleaning up functionality these accelerators no longer use, in order to make the tool's job simpler.

tcal-x (approved):
LGTM

PR #34, opened 2021-03-20
Unroll

alanvgreen:
This would go on top of the macc4 PR.

@tcal-x wdyt? There's only ~0.5M cycles of savings here. How does that balance against the impact on readability of the C code?

Maybe not now, but come back to it later when we need the last few tenths of a percent of performance?


alanvgreen:
hmm... this is failing timing for me now. Will refactor.

PR #37, opened 2021-03-22
Include post processing into Macc4Run1 op

alanvgreen:


tcal-x (approved):
LGTM (admittedly a cursory inspection)

PR #38, opened 2021-03-22
Enable clean exit from simulation ('x' from top menu).

tcal-x:

### DIFF
--- a/common/src/main.c
+++ b/common/src/main.c
@@ -46,7 +46,7 @@ static struct Menu MENU = {
         MENU_ITEM('5', "TFLite Unit Tests", tflite_do_tests),
         MENU_ITEM('6', "Benchmarks", do_benchmarks_menu),
 #ifdef PLATFORM_sim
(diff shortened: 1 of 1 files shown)


common/src/main.c
```diff
@@ -36,6 +45,9 @@ static struct Menu MENU = {
         MENU_ITEM('4', "Performance Counter Tests", perf_test_menu),
         MENU_ITEM('5', "TFLite Unit Tests", tflite_do_tests),
         MENU_ITEM('6', "Benchmarks", do_benchmarks_menu),
+#ifdef PLATFORM_sim
+        MENU_ITEM('x', "Exit (only works in simulation)", exit_sim),
```
alanvgreen, line 49: Consider using Q instead of x to avoid accidental exits.

alanvgreen (approved):
LGTM

common/src/main.c
tcal-x: Done.

PR #52, opened 2021-04-13
Docs5

alanvgreen:


tcal-x:
@alanvgreen , why is HTML better for images here?

tcal-x:
> @alanvgreen , why is HTML better for images here?

Oh, because they link to external images.

PR #56, opened 2021-04-21
Add make as a requirement

PiotrZierhoffer:
This adds `make` as one of the requirements. Does not change much for a regular user, but helps in the CI scenario


tcal-x (approved):
LGTM

PR #60, opened 2021-04-25
hps_soc: add option to use litespi

kgugala:
This PR adds an option to use liteSPI QSPI controller with the HPS platform. It requires a fix from https://github.com/enjoy-digital/litex/pull/891 to work.

Also, I had to change CSR bus width to 32 (liteSPI defines such registers). This can potentially be changed in litespi (if 8 bit CSR bus is a must here).

Signed-off-by: Karol Gugala <kgugala@antmicro.com>

### DIFF
--- a/soc/hps_soc.py
+++ b/soc/hps_soc.py
@@ -29,6 +29,11 @@ from litex.soc.cores.bitbang import I2CMaster
 
 from litex.build.lattice.radiant import radiant_build_args, radiant_build_argdict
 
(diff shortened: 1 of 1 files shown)


tcal-x:
@kgugala , does it make sense to add the litex-hub/litespi submodule in this PR?

kgugala:
it may - this code uses litespi. I'll add it

tcal-x (approved):
We can add the litespi submodule later.

PR #63, opened 2021-04-26
scripts/setup: Improve gcc versioning

### LINKED ISSUE #57: Installing riscv64-unknown-elf-gcc from packages does not guarantee compilation
The docs mention that the RISC-V toolchain has to be downloaded from [here](https://github.com/sifive/freedom-tools/releases/tag/v2020.04.0-Toolchain.Only).

The `scripts/setup` script tries to install it from apt.

This leads to a failure in compilation, complaining about missing cstdio.

In my opinion it would make sense to just point to relevant instructions in case the toolchain is missing.

alanvgreen:
This change prints the URL of the GCC compiler installation
documentation when GCC is not found, or is an unrecognized version.

Currently, setup attempts to install GCC if one is not found. That is
almost always the wrong thing to do.

Fixes #57

### DIFF
--- a/scripts/setup
+++ b/scripts/setup
@@ -51,10 +51,15 @@ if [ ${#missing[@]} -gt 0 ]; then
     fi
 fi
 
-# Warn about different versions
+# Check GCC
+if ! which riscv64-unknown-elf-gcc >/dev/null; then
+    echo "Error: RISCV GCC toolchain not found. Please install one, following the instructions at"
+    echo "https://cfu-playground.readthedocs.io/en/latest/setup-guide.html#step-3-install-riscv-toolchain"
 
-if ! riscv64-unknown-elf-gcc --version | grep ' 8.3.' >/dev/null; then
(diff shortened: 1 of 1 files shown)


danc86 (approved):

tcal-x:
@alanvgreen should we take out the `missing+=(gcc-riscv64-unknown-elf)` logic earlier in the script?

tcal-x (approved):

danc86 (approved):
Not directly related to this PR but FYI: I am using a riscv GCC toolchain built from Sifive's master branch and it gave me 10.2 not 8.3:

```
$ riscv64-unknown-elf-gcc --version
riscv64-unknown-elf-gcc (GCC) 10.2.0
```

I guess they might have dropped the `Sifive` part of the version string because there are no longer any out-of-tree patches?

Debian includes a riscv64-unknown-elf cross toolchain with GCC 10, so we could probably switch to using that at some point -- the issue is just that it doesn't include newlib. Anyway, a problem for another day :-)

PR #65, opened 2021-04-28
Renode pipeline

PiotrZierhoffer:
This creates a sample Renode-based pipeline.

The robot test itself is a stub, but can be expanded to a more sensible scenario.

This pattern can easily be used to do more complex test scenarios in future.


tcal-x (approved):
LGTM

PR #69, opened 2021-05-03
Add Arty spiflash option; add benchmarks when flash is present.

tcal-x:

### DIFF
--- a/common/src/benchmarks.c
+++ b/common/src/benchmarks.c
@@ -21,6 +21,7 @@
 #include "assert.h"
 #include "menu.h"
 #include "perf.h"
+#include "generated/mem.h"
 
 #ifdef PLATFORM_hps
 #define BUF_SIZE (32 * 1024)  // must be at least 1024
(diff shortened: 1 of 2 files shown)


danc86 (approved):
Neat. I guess you are using this to narrow down the performance problems with Litespi that you were seeing?

I tried it on my Arty and got some surprising results:

```
Running flash sequential loads benchmark
Hello, Flash Load! (SPI flash at 0x80000000, size 16384kB)
Val:-154088964  Cycles: 3877489   Cycles/load: 59
```

```
Running flash STRIDED loads benchmark
Hello, STRIDED Flash Load! (SPI flash at 0x80000000, size 16384kB)
Val:1987943126  Cycles: 675152657   Cycles/load: 10302
```

170 times slower with the strided read... seems a bit surprising. I guess the strided read pattern is defeating Litespi's read-ahead mechanism, similar to how a random access pattern will often do?

tcal-x:
> Neat. I guess you are using this to narrow down the performance problems with Litespi that you were seeing?
> 
> I tried it on my Arty and got some surprising results:
> 
> ```
> Running flash sequential loads benchmark
> Hello, Flash Load! (SPI flash at 0x80000000, size 16384kB)
> Val:-154088964  Cycles: 3877489   Cycles/load: 59
> ```
> 
> ```
> Running flash STRIDED loads benchmark
> Hello, STRIDED Flash Load! (SPI flash at 0x80000000, size 16384kB)
> Val:1987943126  Cycles: 675152657   Cycles/load: 10302
> ```
> 
> 170 times slower with the strided read... seems a bit surprising. I guess the strided read pattern is defeating Litespi's read-ahead mechanism, similar to how a random access pattern will often do?

@danc86 , yes, that's exactly what I get!

With this merged, Karol can take a look.

I see something in the litespi gateware that has a 10000 cycle timeout, with "FIXME" next to it;   that, I think, would explain the 302 (a reasonable amount of cycles) + 10000 (?)

tcal-x:
Hi @kgugala , I was playing around with litespi on Arty in this repo since we're already doing some benchmarking here (see the file changes in this PR).   Is the extra 10000 cycles/iteration expected for non-sequential loads?   The "strided" access pattern is accessing every other 32b int, so buf[0], buf[2], buf[4], etc. but actually it didn't make much difference -- as long as you did anything other than perfectly sequential accesses, the cycles/iteration was 10302.

I found a "10000" constant here: https://github.com/litex-hub/litespi/blob/master/litespi/core/mmap.py#L106-L107 , and when I changed it to 5000, the cycles/iteration measured here then became 5302.  

Is this something to change in mmap.py, in how we instantiate litespi, or do we need to change how we access this memory region?   If a complete cache line is read, then a randomly different cache line is read, will we get better behavior?    

To reproduce in this repo, attach your Arty A7 35T board,
* hand edit soc/arty.py to force the mapped flash option to True
* `cd` to `proj/proj_template_v`
* `make clean`
* `make prog`
* `make -j8 software`
* `make load`
* select 6 (benchmarks), then 'f' for sequential flash loads, 'g' for strided flash loads

mithro:
FYI - @kgugala

mithro:
@tcal-x -- That just outright looks like a bug / unfinished code....

kgugala:
another thing I see is that we're still using default clock prescaler of 9 https://github.com/google/CFU-Playground/blob/main/soc/hps_soc.py#L144 - this can be lowered by passing `default_divisior` to the PHY constructor https://github.com/litex-hub/litespi/blob/master/litespi/phy/generic.py#L358

kgugala:
I tweaked the settings a bit and got:

```
Running flash STRIDED loads benchmark
Hello, STRIDED Flash Load! (SPI flash at 0x80000000, size 16384kB)
Val:284099572  Cycles: 19925004   Cycles/load: 304
```

PR #71, opened 2021-05-03
HPS updates: allow alt synth tool; Renode files.

tcal-x:
* Add "USE_LSE=1" and "USE_YOSYS=1" options for gateware build
  - these still use Radiant place and route
* Add Renode platform and script files for HPS
  - only in proj_template_v; I need to centralized these...

### DIFF
--- a/proj/proj_template_v/renode/hps.repl
+++ b/proj/proj_template_v/renode/hps.repl
@@ -1,0 +1,62 @@
+
+cpu: CPU.VexRiscv @ sysbus
+    cpuType: "rv32im"
+    init:
+        RegisterCustomCSR "BPM" 0xB04  User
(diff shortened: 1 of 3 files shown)


danc86 (approved):
Oooh a Renode definition is handy -- we were just talking yesterday about the possibility of using Renode to run the HPS SoC application ROM but I had no idea how to go about it. Looks like you have done the hard part already :-)

tcal-x:
@danc86 , Making the .repl file is basically just copying memory region locations and CSRs (timer, uart) from the csr.csv and maybe other files generated by LiteX.   If I were smart, I'd automate the creation of the .repl file; I think such scripts already exist.

danc86:
Oh I see, good to know, thanks.

PR #74, opened 2021-05-05
mnv2_first tests

PiotrZierhoffer:
This PR adds tests for both HPS and LiteX for mnv2_first project.

These tests will, eventually, be quite similar for all platforms. Right now they differ in one test case.
I would suggest keeping them separate for now and try to unify when we have more actual cases.

GitHub Actions workflow, on the other hand, is something that should be definitely unified.

I took the liberty of using @mithro 's actions-includes. This really solves all the issues with code reuse - otherwise each job is mostly copy-paste.
It does, however, require some additional consideration:

1. I used the `.github/workflows-src/Makefile` from SymbiFlow/fasm. It has, obviously, `SymbiFlow Authors` copyrights. Is this an issue?
2. The `.github/workflows/test-projects.yml` is autogenerated. It requires running `make update` in `.github/workflows-src` whenever the workflow source files change. I don't think it's a big issue, when we compare to the number of issues it solves.

So @tcal-x if you consider these changes to be ok, then I think this is mergable. Otherwise I can rely on plain old copy-paste to create the workflow.

I also add Robot output files to gitignore.

Quick PR summary:

* Two robot files, for hps and litex-vexriscv-zephyr, for the mnv2_first project
* .github/includes/actions/run-in-renode/action.yml - the base for Renode tests. It's generally the same action we had there before for proj_template, but now it's parametrized with project path, test file path and build flags
* .github/workflows-src - template files for workflows going into .github/workflows, along with their build system
* .github/workflows/test-projects.yml - autogenerated workflow file

### DIFF
--- a/.github/workflows/run-in-renode.yml
+++ b/.github/workflows/run-in-renode.yml
@@ -1,28 +1,52 @@
+
+# !! WARNING !!
+# Do not modify this file directly!
+# !! WARNING !!
+#
+# It is generated from: ../workflows-src/run-in-renode.yml
+# using the script from https://github.com/mithro/actions-includes@main
+
 name: Run in Renode
 on: [push, pull_request, workflow_dispatch]
 jobs:
-  run-in-renode:
+  proj_template:
     runs-on: ubuntu-20.04
     steps:
-      - run: wget -O- https://static.dev.sifive.com/dev-tools/riscv64-unknown-elf-gcc-8.3.0-2020.04.1-x86_64-linux-ubuntu14.tar.gz | tar -xzC /opt
-      - run: echo "/opt/riscv64-unknown-elf-gcc-8.3.0-2020.04.1-x86_64-linux-ubuntu14/bin" >> $GITHUB_PATH
-      - uses: actions/checkout@v2
-      - uses: actions/setup-python@v2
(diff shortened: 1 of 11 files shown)


.github/workflows/run-in-renode.yml
```diff
@@ -1,28 +1,52 @@
...
-          tests-to-run: 'proj/proj_template/renode/litex-vexriscv-tflite.robot'
-      - name: Archive results
-        uses: actions/upload-artifact@v2
-        with:
-          name: test-results
-          path: |
-            report.html
-            log.html
-            robot_output.xml
+    - name: ⏰ 🛂 📕 - Checking workflow expansion is up to date
```
tcal-x: @PiotrZierhoffer does this check that I didn't forget to do the `make update` in `.github/workflows-src/` ?

PiotrZierhoffer: That's correct. Actually, this step is added automatically by actions-include

tcal-x: > That's correct. Actually, this step is added automatically by actions-include

That's nifty!

tcal-x (approved):
LGTM

PR #75, opened 2021-05-06
Add GCC dependency

PiotrZierhoffer:
This is required to build flashrom.

### DIFF
--- a/scripts/setup
+++ b/scripts/setup
@@ -7,6 +7,9 @@ git submodule update --init
 # Verify dependencies that can be installed with apt (if available)
 missing=()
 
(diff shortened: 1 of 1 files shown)


tcal-x (commented):
Hi @PiotrZierhoffer  -- what is this needed for?   It's to build something that runs on the host, right?

PiotrZierhoffer:
The setup script builds flashrom, and it requires gcc to be installed.

tcal-x:
> The setup script builds flashrom, and it requires gcc to be installed.

Ah now I see your earlier comment.

tcal-x (approved):
LGTM

PR #76, opened 2021-05-06
Add a simple Dockerfile with all dependencies

PiotrZierhoffer:
I'm not really sure if this should be merged. I created this Dockerfile and I use it for my development. Maybe someone will find it useful.

But I'm ok with rejecting this PR if you don't see it fitting in this repo.

Build the dockerfile with `cd scripts; docker build -t cfu-playground .`

Run with `docker run -ti cfu-playground`

### DIFF
--- a/scripts/Dockerfile
+++ b/scripts/Dockerfile
@@ -1,0 +1,16 @@
+FROM debian:testing
+
+ENV RISCV_DIR=/toolchain/riscv64-unknown-elf-gcc-8.3.0-2020.04.1-x86_64-linux-ubuntu14/bin
+ENV PATH="/renode:${RISCV_DIR}:${PATH}"
+ENV WORKDIR=/CFU-Playground
+
+RUN apt update -y && apt install -y wget git python3-pip make gcc openocd yosys expect ccache verilator libevent-dev libjson-c-dev libusb-1.0-0-dev libftdi1-dev vim && rm -rf /var/lib/apt/lists/*
+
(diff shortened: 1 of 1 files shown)


scripts/Dockerfile
```diff
@@ -0,0 +1,16 @@
+FROM debian:testing
+
+ENV RISCV_DIR=/toolchain/riscv64-unknown-elf-gcc-8.3.0-2020.04.1-x86_64-linux-ubuntu14/bin
```
irvifa, line 3: Won't this only needed during the build time? In that time probably we can change it to `ARG`
```suggestion
ARG RISCV_DIR=/toolchain/riscv64-unknown-elf-gcc-8.3.0-2020.04.1-x86_64-linux-ubuntu14/bin
```

scripts/Dockerfile
```diff
@@ -0,0 +1,16 @@
...
+ENV PATH="/renode:${RISCV_DIR}:${PATH}"
+ENV WORKDIR=/CFU-Playground
```
irvifa, line 5: Won't this only needed during the build time? In that time probably we can change it to `ARG`
```suggestion
ARG WORKDIR=/CFU-Playground
```

scripts/Dockerfile
PiotrZierhoffer: This variable is listed by the README, but I'm not fully sure it's needed. I will double check it.

tcal-x: Correct, RISCV_DIR is not needed or used.

PR #77, opened 2021-05-06
Fix some stuff that hadn't been tested in a while.

tcal-x:
* remove "--no-crc" from expects script
* fix scanprof.py regex for new profiler output


alanvgreen:
LGTM

PR #78, opened 2021-05-06
Pass --synth-mode option to Radiant tools (lse, yosys).

tcal-x:
Default is Synplify.
Specify 'yosys' to get Yosys,
or 'lse' to get (Radiant) LSE.

### DIFF
--- a/soc/hps_soc.py
+++ b/soc/hps_soc.py
@@ -211,6 +211,8 @@ def main():
                         help="Whether to do a full build, including the bitstream")
     parser.add_argument("--toolchain", default="radiant",
                         help="Which toolchain to use, radiant (default) or oxide")
+    parser.add_argument("--synth_mode", default="radiant",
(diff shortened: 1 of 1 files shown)


tcal-x:
FYI  `--synth-mode` is used here: https://github.com/enjoy-digital/litex/blob/master/litex/build/lattice/radiant.py

danc86 (approved):
The commit message mentions Synplicity and I got scared for a second because that's yet another thing I had not heard of... but then I realised it is just a typo for Synplify, part of the Radiant toolchain. :-)

tcal-x:
> The commit message mentions Synplicity and I got scared for a second because that's yet another thing I had not heard of... but then I realised it is just a typo for Synplify, part of the Radiant toolchain. :-)

Oh yes, I always mix them up.  Let me make sure the message in the code is right.

Issue #84, opened 2021-05-14
Can GitHub actions be made conditional based on what was changed?


tcal-x:
E.g. If a PR only impacts files under docs/, then the RTD action should run, but not the build/Renode actions.

tcal-x:
Fixed with PR #88.

PR #88, opened 2021-05-19
Do not run workflows on unrelated path changes

PiotrZierhoffer:
This is to address #84 .

I'm not yet sure if we should have a filter for gh action yml changes, investigating.


tcal-x:
> I'm not yet sure if we should have a filter for gh action yml changes, investigating.

If I'm right, the GHA directory is **not** ignored for triggering the software/build tests, which is what we'd want, so it's good.

tcal-x (approved):
LGTM

PR #93, opened 2021-05-26
Solving example_cfu_v and add qmtech wukong board

akioolin:
there two parts update.

1. acording to TCal's guide, add rst input signal to solve synthesis fail.
2. add qmtech wukong board support.


akioolin:
@googlebot I signed it!

PR #96, opened 2021-05-26
add rst into proj/example_cfu_v/cfu.v

akioolin:
follow TCal-X, add rst signal into proj/example_cfu_v/cfu.v

### DIFF
--- a/proj/example_cfu_v/cfu.v
+++ b/proj/example_cfu_v/cfu.v
@@ -24,6 +24,7 @@ module Cfu (
   input               io_bus_rsp_ready,
   output              io_bus_rsp_payload_response_ok,
   output     [31:0]   io_bus_rsp_payload_outputs_0,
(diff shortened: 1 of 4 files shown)


tcal-x (approved):
LGTM

tcal-x:
Thanks @akioolin , it is merged.

I should mention that we hope to generalize the board handling so that we don't need a special board.py and board.mk file for each board, at which point these changes will be discarded.

akioolin:
> Thanks @akioolin , it is merged.
> 
> I should mention that we hope to generalize the board handling so that we don't need a special board.py and board.mk file for each board, at which point these changes will be discarded.


Hi, @tcal-x :

I see. for CFU, it should be a internal IO function block. the IO could be control signal, status signal, memory bus and interrupt. for board specified setting is the very beginning state.

BR, Akio

PR #100, opened 2021-05-27
change variable setting and range in  do_exercise_cfu_op0()

akioolin:
change variable setting from signed to unsigned and for range in 
do_exercise_cfu_op0() of proj/proj_template/src/proj_menu.c


tcal-x (approved):
LGTM

Issue #86, opened 2021-05-17
Simplify the top LiteX scripts now that CFU integration is upstreamed.


tcal-x:
There is now a `--cpu-cfu` option in LiteX's SocCore, and the +cfu variant is understood by cores/cpu/vexriscv/core.py, so we no longer need to use an external variant and we no longer need the VexRiscv Verilog "wrapper".

tcal-x:
Specifically, soc/arty.py, soc/nexys_video.py, and soc/hps_soc.py can be simplified.

PiotrZierhoffer:
This is ongoing on our side

PiotrZierhoffer:
@tcal-x I guess we can close this one?

tcal-x:
Yes, thanks Piotr and team!

PR #101, opened 2021-05-27
Simplify top LiteX scripts after LiteX CFU integration

robertszczepanski:
This is follow-up to #86. 
It simplifies `arty`, `nexys_video`, `hps_soc` and `qmtech_wukong` top scripts since there is no need to use external variants anymore. 
It also modifies `cfu.v` since LiteX doesn't expect `io_bus_` prefix. 
In case of hps soc I decided to leave support for `slim` variant as it is and add support for LiteX CFU integration for other variants that are supported by LiteX.
I've also removed redundant `$(LITEX_ARGS)` from `$(BOARD_RUN) $(LITEX_ARGS)` in `soc/*.mk` files because it is already in `BOARD_RUN` e.g. `ARTY_RUN:=  MAKEFLAGS=-j8 $(PYRUN) ./arty.py $(LITEX_ARGS)`. It did make `$(LITEX_ARGS)` doubled.

It has been tested on Arty A7 35-T and Nexys Video Board.

### DIFF
--- a/proj/proj_template_v/cfu.v
+++ b/proj/proj_template_v/cfu.v
@@ -15,30 +15,30 @@
 
 
 module Cfu (
-  input               io_bus_cmd_valid,
-  output              io_bus_cmd_ready,
-  input      [2:0]    io_bus_cmd_payload_function_id,
-  input      [31:0]   io_bus_cmd_payload_inputs_0,
-  input      [31:0]   io_bus_cmd_payload_inputs_1,
-  output              io_bus_rsp_valid,
-  input               io_bus_rsp_ready,
-  output              io_bus_rsp_payload_response_ok,
-  output     [31:0]   io_bus_rsp_payload_outputs_0,
-  input               rst,
+  input               cmd_valid,
+  output              cmd_ready,
+  input      [2:0]    cmd_payload_function_id,
+  input      [31:0]   cmd_payload_inputs_0,
+  input      [31:0]   cmd_payload_inputs_1,
+  output              rsp_valid,
+  input               rsp_ready,
+  output              rsp_payload_response_ok,
+  output     [31:0]   rsp_payload_outputs_0,
+  input               reset,
   input               clk
 );
 
   // Trivial handshaking for a combinational CFU
-  assign io_bus_rsp_valid = io_bus_cmd_valid;
-  assign io_bus_cmd_ready = io_bus_rsp_ready;
-  assign io_bus_rsp_payload_response_ok = 1'b1;
+  assign rsp_valid = cmd_valid;
+  assign cmd_ready = rsp_ready;
+  assign rsp_payload_response_ok = 1'b1;
 
   //
   // select output -- note that we're not fully decoding the 3 function_id bits
   //
-  assign io_bus_rsp_payload_outputs_0 = io_bus_cmd_payload_function_id[0] ? 
-                                           io_bus_cmd_payload_inputs_1 :
-                                           io_bus_cmd_payload_inputs_0 ;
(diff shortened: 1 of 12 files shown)


tcal-x:
@robertszczepanski @PiotrZierhoffer -- thanks for catching the double `$LITEX_ARGS`, and also catching the port name changes (I forgot to warn you about that).

I expect there will be some additional small changes needed for the nMigen-based projects --- I'm hoping it's just one change under python/nmigen_cfu/, and not something that needs changing in every proj/*/.

tcal-x:
This also points out that we could add some more tests, since all of the tests pass even with the port name mismatch I mention above, which we would see in the mnv2_first project.   I think issues like that would be caught if CI tests included Verilator simulation.   This path already exists in the form of full-SoC simulation (`make PLATFORM=sim load`), which of course is slow.   But even just building the Verilated simulation should catch it?   Currently building and running the simulation are not separate tasks, and the running is interactive.   And since it's not Renode-based, the test would use `expect` and not `robot`.    Or, we could wait for Renode-Verilator co-sim for such a test.   

I'll open a separate issue.

robertszczepanski:
@tcal-x I've applied some changes in python/nmigen. I built `proj/example_cfu` and loaded on Arty. It looks like working well so I hope I didn't miss anything.

More tests would be great btw :)

python/nmigen_cfu/cfu.py
```diff
@@ -157,15 +157,15 @@ class Cfu(SimpleElaboratable):
...
-        input      [19:0]   io_bus_cmd_payload_function_id,
-        input      [31:0]   io_bus_cmd_payload_inputs_0,
-        input      [31:0]   io_bus_cmd_payload_inputs_1,
-        output              io_bus_rsp_valid,
-        input               io_bus_rsp_ready,
-        output              io_bus_rsp_payload_response_ok,
-        output     [31:0]   io_bus_rsp_payload_outputs_0,
+        input               cmd_valid,
+        output              cmd_ready,
+        input      [9:0]    cmd_payload_function_id,
```
tcal-x, line 162: Good catch!

tcal-x (approved):
LGTM

I will do a bit of manual testing before merging.

proj/proj_template_v/cfu.v
```diff
@@ -15,30 +15,30 @@
...
+  input               cmd_valid,
+  output              cmd_ready,
+  input      [2:0]    cmd_payload_function_id,
+  input      [31:0]   cmd_payload_inputs_0,
+  input      [31:0]   cmd_payload_inputs_1,
+  output              rsp_valid,
+  input               rsp_ready,
+  output              rsp_payload_response_ok,
+  output     [31:0]   rsp_payload_outputs_0,
   input               rst,
```
tcal-x, line 27: I found that 'rst' needs to be updated to 'reset'.  @robertszczepanski , can you make that fix, or check the box that gives me permission to make edits on your branch?

tcal-x:
Also, I will need to update the VexRiscv Verilog submodule to not use the wrapper methodology as part of this PR.   We can just ignore the wrapper itself, but part of the scheme is renaming the actual VexRiscv core to be "VexRiscv_inner", so we need to stop doing that renaming.

robertszczepanski:
I changed `rst` to `reset`. I also did change the `VexRiscv_inner` name locally but I forgot to mention that it must be done.

If your manual tests didn't expose any other deficiencies I guess it's good now.

tcal-x:
@robertszczepanski thanks for the updates!

tcal-x:
Closing as these commits were merged with PR #103.   Thanks again @robertszczepanski .

PR #104, opened 2021-06-03
Add models from TFLM examples.

JosephBushagour:
Two other models from the [tflite-micro examples](https://github.com/tensorflow/tflite-micro/tree/main/tensorflow/lite/micro/examples) (micro_speech and magic_wand) are moved into the available models for the CFU-Playground. They include golden tests as well. The magic_wand model is not quantized because the TFLM example doesn't have it quantized. In the future, a quantized version could be added with the benefit of a significant speed improvement.

The `models.c` file is reworked to be opt-in for each model (based on defined identifiers). The project template `Makefiles` include all TFLM examples by default. Older projects have the `pdti8` model included (as was default). The menu will also warn users if no models are included.

The bulk of files changed and lines modified are directly from the tflite-micro examples and are located in `third_party/tflm_gen/tensorflow/lite/micro/examples/`.


alanvgreen:
LGTM. Thanks for being thorough.

tcal-x (approved):
LGTM

Issue #85, opened 2021-05-17
Enable any LiteX-supported board to be used.


tcal-x:
CFU-Playground currently supports Arty 35T and Nexys Video boards.   The difference between them is very small, but the current setup requires duplication of at least a couple of files in the soc/ directory (arty.mk, arty.py, nexys_video.mk, nexys_video.py).   Instead, the board should be a parameter so that just one .mk and one .py file are needed.

Of course, not all LiteX-supported boards have the fabric resources to handle a "full" VexRiscv variant, or the memory to handle both the loaded executable and the necessary work space (e.g. for the tensor arena), so we may want to perform some up-front checks.

PiotrZierhoffer:
@tcal-x should we close this issue?

PR #105, opened 2021-06-04
Add support for all LiteX Boards targets

robertszczepanski:
Follow-up to #85.

This adds `common_soc` that can take any LiteX Boards targets name as `TARGET` flag e.g. `make bitstream TARGET=nexys_video_board`. By default it's set to `digilent_arty`. 
Toolchain is now taken from board's platform file e.g. https://github.com/litex-hub/litex-boards/blob/master/litex_boards/platforms/digilent_arty.py#L338. That's why there's `USE_VIVADO` to force Vivado toolchain.
Readme has been updated to contain info about more boards than just the Arty. There is additional information about possible make flags.
I've removed not needed top scripts for `arty`, `nexys_video` and `qmtech_wukong` boards since you can pass these as `TARGET` flag.

There's also small fix of `PLATFORM` check. It didn't really work for me before that.


tcal-x:
Thank you @robertszczepanski , I'm really excited about this update.   And yes, the Makefile flags were a bit of a mess.  I'll do some testing later today.

soc/common_soc.mk
```diff
@@ -69,11 +74,11 @@ check-timing:
...
 else
 check-timing:
-	@echo Timing check not performed for SymbiFlow.
+	@echo Timing check performed only for Vivado.
 endif
 
 prog: $(BITSTREAM) check-timing
-	@echo Loading bitstream onto Arty
+	@echo Loading bitstream onto board
 	$(ARTY_RUN) --no-compile-software --load
```
tcal-x, line 82: Missed one -- $(ARTY_RUN)

robertszczepanski: Thanks! Done.

soc/common_soc.mk
```diff
@@ -35,29 +35,34 @@ endif
 PROJ_DIR:=  $(CFU_ROOT)/proj/$(PROJ)
 CFU_V:=     $(PROJ_DIR)/cfu.v
 CFU_ARGS:=  --cpu-cfu $(CFU_V)
+TARGET_ARGS:= --target $(TARGET)
 
-SOC_NAME:=  nexys_video.$(PROJ)
+SOC_NAME:=  common_soc.$(PROJ)
```
tcal-x, line 40: Can we change this to `SOC_NAME:= $(TARGET).$(PROJ)` ?

robertszczepanski: Sure, I changed that but to make it work I had to add [a statement](https://github.com/google/CFU-Playground/pull/105/files#diff-13f94e0336e4f0044b0809865aa44e3d111cb67a44b948d2a98f0268984eb5f6R63) that makes sure `$(TARGET)` flag is properly set. It's necessary for platforms other than `common_soc`

tcal-x (approved):
LGTM

PR #106, opened 2021-06-04
Replace fp32 magic wand model with quantized version.

JosephBushagour:
The previous model in `common/src/models/magic_wand` was directly used from the TFLM examples. This model was not quantized however, and therefore performed slowly in this environment as all the operations were performed on 32 bit floats. This PR replaces this model with a quantized version of the model that performs faster, resulting in a 75% decrease in cycles spent (12M -> 3M cycles).

This PR also includes a change which prints approximate floating point results from the model (because we can't print floats with `%f` in the playground) rather than just the hexadecimal encoding of the result.


common/src/models/magic_wand/magic_wand.c
```diff
@@ -53,36 +53,44 @@ MagicWandResult magic_wand_classify() {
   };
 }
 
+static void print_magic_wand_result(const char* prefix, MagicWandResult res) {
+  printf("%s-- wing: 0x%x, ring: 0x%x, slope: 0x%x, negative: 0x%x\n", prefix,
+         res.wing_score, res.ring_score, res.slope_score, res.negative_score);
+  printf(
+      "%s approx-- wing: 0.%06ld ring: 0.%06ld, slope: 0.%06ld, negative: "
+      "0.%06ld\n",
+      prefix, (int)(*(float*)&res.wing_score * 1000000),
```
alanvgreen, line 62: It looks like it would be more convenient if the MagicWandResult member variables were floats. Any reason not to do that?

common/src/models/magic_wand/magic_wand.c
```diff
@@ -53,36 +53,44 @@ MagicWandResult magic_wand_classify() {
...
 
 #define NUM_GOLDEN 3
 
 static void do_golden_tests() {
+  // Dequantization leads to simpler floats.
   static MagicWandResult magic_wand_golden_results[NUM_GOLDEN] = {
-      {0x3d32c77d, 0x3ee047bc, 0x3db0289e, 0x3edd552d},
-      {0x3d20dcb3, 0x3f4d08cb, 0x3cfb854a, 0x3e043500},
-      {0x39df0176, 0x381045eb, 0x3f742eda, 0x3d3b305e},
+      {0x3d200000, 0x3ee00000, 0x3db00000, 0x3ee00000},
```
alanvgreen, line 91: Ah.. this might be the reason for not storing members as floats. I suggest making a MagicWandResult constructor which  takes four uint32_t parameters and casts them to floats.

This change would not have to block this PR - it could be done in a folllow up.

alanvgreen (commented):
@danc86 What would be the best way to have printf() handle floats? 

I'll let @tcal-x approve this PR.

common/src/models/magic_wand/magic_wand.c
JosephBushagour: Yes this was the initial reason for using `uint32_t` members. I think making the change to `float` members is smart, but would we still want to print out the hexadecimal representation of the bits in our tests? I would think yes because without the hexadecimal bit representation we could run into issues with failed tests where the offending inequality is not visible with just 6 decimal places.

tcal-x (approved):
LGTM

danc86:
The `printf()` in Litex libbase has optional float support, the Makefile produces `libbase-nofloat.a` without float support and `libbase.a` with float support compiled in. This may be enough to get working `%f` in `printf()`:

```
diff --git a/common/Makefile b/common/Makefile
index 88c3d4b..d204a60 100644
--- a/common/Makefile
+++ b/common/Makefile
@@ -100,7 +100,7 @@ LFLAGS     := \
        $(CXXFLAGS) \
        -L$(LD_DIR) \
        -L$(GEN_LD_DIR) \
-       -L$(SOC_SOFTWARE_DIR)/libbase -lbase-nofloat \
+       -L$(SOC_SOFTWARE_DIR)/libbase -lbase \
        -L$(SOC_SOFTWARE_DIR)/libcompiler_rt -lcompiler_rt \
        -lm \
        -nostartfiles \
```

I can test it out and send a PR, if `%f` will be helpful for this model.

PR #111, opened 2021-06-08
Add the MLCommons Tiny benchmark models and tests.

JosephBushagour:
The reference models from the [MLCommons Tiny Benchmark](https://github.com/mlcommons/tiny) are included in the possible models a user can opt-in to. These models are located in a new folder: `third_party/mlcommons/tiny/...`. This PR does not implement the actual benchmark, as that requires some resources that aren't too easy to provide in the CFU-Playground (complete control of the serial port, predefined submitter-implemented functions, etc.).

Each model has associated test cases and expected outputs (golden tests). The anomaly detection model has an output tensor of 640 elements -- to avoided checking each of these elements for the correct values the output tensor is xor reduced to a 32 bit checksum which is used for comparison. The inputs for each model are quantized if needed, this simplifies the model testing in the playground and avoids ever needed to do floating point arithmetic. 

More information about the included models can be found in the README file located at `common/src/models/mlcommons_tiny_v01/README.md`.


tcal-x (approved):
LGTM

PR #113, opened 2021-06-11
In nMigen Cfu, tie implicit "rst" to explicit "reset".

tcal-x:
This is a suboptimal fix because the generated Verilog Cfu module will have both "rst" and "reset" ports.   This PR ties them together, so that even though only "reset" is driven by sys_rst, it gets passed to the "rst" net inside the Cfu.


danc86:
It looks like this solves the issue I was seeing where Yosys optimised the whole design away, judging by the much longer pnr time and resource utilisation when I apply this patch. The design still doesn't work but I think that's a separate issue in my design.

Issue #112, opened 2021-06-11
Versioning TfLiteMicro


alanvgreen:
We should be able to pick which version of TfLiteMicro to use.

In particular, eing able to point to a particular hash of https://github.com/tensorflow/tflite-micro could be helpful.

PR #118, opened 2021-06-16
Update tflite-micro version

alanvgreen:
Updates tflite to use the new tflite-micro git repository. Currently, still only supports a single version. Tested against proj_template, avg_pdti8, proj_accel_1 and mnv2_first with Arty.

#112


tcal-x:
tflmyay
Yay!

tcal-x (approved):
LGTM!

akioolin:
Hi, @alanvgreen  :

After some testing with this update,  I met a problem. when compile tflm library, it complained no found to "qmtech_wukong_makefile.inc" wihch the board I used is wukong board.

Would you like to share how to solve this problem? or just simple copy the mcu_riscv_makefile.inc to digilet_arty_makefile.inc? 

Many thanks. 

BR, Akio

tcal-x:
@akioolin thank you for the report.   There has been a lot of restructuring; first, the board-specific .py/.mk files were eliminated in favor of common_soc.py/mk.    More recently, @JosephBushagour  has added the ability to tailor a board-specific workflow when necessary.    We'll try to figure out where things went wrong for the qmtech_wukong.

tcal-x:
Hi @akioolin , I think I see the issue.   It is due to the interaction of this PR and PR #105.    #105 added the use of TARGET to specify which LiteX board we're using.   This PR causes more of the TFLite-Micro system to be used for generating the build/src/tensorflow/lite/micro/ directory, and *that* also uses TARGET but for a different purpose.   Hopefully the solution is as easy as just unsetting TARGET before calling the TFLM generation.

PR #119, opened 2021-06-17
hps: also expose command line options for the Oxide toolchain

danc86:
The --nextpnr-timingstrict and --nextpnr-seed options are of particular
interest.

### DIFF
--- a/soc/hps_soc.py
+++ b/soc/hps_soc.py
@@ -27,6 +27,7 @@ from litex.soc.cores.spi_flash import SpiFlash
 from litex.soc.cores.bitbang import I2CMaster
 
 from litex.build.lattice.radiant import radiant_build_args, radiant_build_argdict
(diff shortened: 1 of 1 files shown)


tcal-x (approved):
Thanks Dan!   I had similar changes in my sandbox, but in common_soc.py.   In that case, I ran into an issue of the same option being added for different tool chains (`--synth-mode` for both Vivado and Radiant).    I was able to get past that by adding `conflict_handler='resolve'` to the ArgumentParser constructor.

Here we don't have that issue.

LGTM

danc86:
Oh, good to know. I guess it's inevitable we will start to have conflicting options as these scripts continue to grow more and more options.

I'm looking forward to a point where we can just use the yosys/symbiflow toolchain everywhere and take out all the extra complexity for choosing vendor toolchains!

PR #121, opened 2021-06-17
Use dcache metrics CSRs in benchmarks

fkokosinski:
This PR introduces displaying of additional dcache-related matrics from CPU's CSRs in benchmarks:

```
Running sequential stores benchmark
Hello, Store!
Val:28  Cycles: 11389084   Cycles/store: 10
[Dcache accesses] Before: 3223075888, After: 3224127310, Diff: 1051422
[Dcache refills]  Before: 952539, After: 952545, Diff: 6
[Dcache stalls]   Before: 89447457, After: 96635863, Diff: 7188406
```

### DIFF
--- a/common/src/benchmarks.c
+++ b/common/src/benchmarks.c
@@ -17,9 +17,11 @@
 #include "benchmarks.h"
 
 #include <stdio.h>
+#include <stdint.h>
 
 #include "assert.h"
 #include "menu.h"
+#include "metrics.h"
 #include "perf.h"
 #include "generated/mem.h"
 
@@ -39,6 +41,8 @@ static void __attribute__((noinline)) do_loads_cached(void) {
   for (int j = 0; j < 1024; ++j) {  // warmup
     acc += buf[j];
   }
+
+  DCACHE_SETUP_METRICS;
   int start_time = perf_get_mcycle();
   for (int i = 0; i < 1024; ++i) {
     for (int j = 0; j < 1024; ++j) {
@@ -47,9 +51,13 @@ static void __attribute__((noinline)) do_loads_cached(void) {
     buf[i] = acc;  // inhibit optimization
   }
   int end_time = perf_get_mcycle();
+  DCACHE_PRINT_METRICS;
+
   int total_iters = 1024 * 1024;
-  printf("Val:%d  Cycles: %d   Cycles/load: %d\n\n\n", acc,
+  printf("Val:%d  Cycles: %d   Cycles/load: %d\n", acc,
          end_time - start_time, (end_time - start_time) / total_iters);
+
+  printf("\n\n");
 }
 
 static void __attribute__((noinline)) do_loads_strided(void) {
@@ -57,6 +65,8 @@ static void __attribute__((noinline)) do_loads_strided(void) {
   int buf[BUF_SIZE];
   int acc = 0;
   int start_time = perf_get_mcycle();
+
+  DCACHE_SETUP_METRICS;
   for (int i = 0; i < 64; ++i) {
     // 32 bytes per cache line, so stride by 8 4-byte ints
     for (int j = 0; j < BUF_SIZE; j += 8) {
@@ -65,9 +75,13 @@ static void __attribute__((noinline)) do_loads_strided(void) {
     buf[i] = acc;  // inhibit optimization
   }
   int end_time = perf_get_mcycle();
+  DCACHE_PRINT_METRICS;
+
   int total_iters = 64 * BUF_SIZE / 8;
-  printf("Val:%d  Cycles: %d   Cycles/load: %d\n\n\n", acc,
+  printf("Val:%d  Cycles: %d   Cycles/load: %d\n", acc,
          end_time - start_time, (end_time - start_time) / total_iters);
+
+  printf("\n\n");
 }
 
 #ifdef SPIFLASH_BASE
@@ -76,30 +90,42 @@ static void __attribute__((noinline)) do_flash_loads(void) {
   printf("Hello, Flash Load! (SPI flash at %p, size %dkB)\n", SPIFLASH_BASE, (SPIFLASH_SIZE/1024));
   int*  buf = (int*)((void*)(SPIFLASH_BASE));
   int acc = 0;
+
+  DCACHE_SETUP_METRICS;
   unsigned int start_time = perf_get_mcycle();
   int total_iters = 64*1024;
   for (int j = 0; j<total_iters; j++) {
     acc += buf[j];
   }
   unsigned int end_time = perf_get_mcycle();
+  DCACHE_PRINT_METRICS;
+
   unsigned int delta_cycles = end_time - start_time;
-  printf("Val:%d  Cycles: %u   Cycles/load: %u\n\n\n", acc,
+  printf("Val:%d  Cycles: %u   Cycles/load: %u\n", acc,
          delta_cycles, delta_cycles / total_iters);
+
+  printf("\n\n");
 }
 
 static void __attribute__((noinline)) do_flash_loads_strided(void) {
   printf("Hello, STRIDED Flash Load! (SPI flash at %p, size %dkB)\n", SPIFLASH_BASE, (SPIFLASH_SIZE/1024));
   int*  buf = (int*)((void*)(SPIFLASH_BASE));
   int acc = 0;
+
+  DCACHE_SETUP_METRICS;
   unsigned int start_time = perf_get_mcycle();
   int total_iters = 64*1024;
   for (int j = 0; j<total_iters; j++) {
     acc += buf[j*2];    // 2x stride
   }
   unsigned int end_time = perf_get_mcycle();
+  DCACHE_PRINT_METRICS;
+
   unsigned int delta_cycles = end_time - start_time;
-  printf("Val:%d  Cycles: %u   Cycles/load: %u\n\n\n", acc,
+  printf("Val:%d  Cycles: %u   Cycles/load: %u\n", acc,
          delta_cycles, delta_cycles / total_iters);
+
+  printf("\n\n");
 }
 #endif
 
@@ -107,6 +133,8 @@ static void __attribute__((noinline)) do_loads(void) {
   puts("Hello, Load!\n");
   int buf[BUF_SIZE];
   int acc = 0;
+
+  DCACHE_SETUP_METRICS;
   int start_time = perf_get_mcycle();
   for (int i = 0; i < 8; ++i) {
     for (int j = 0; j < BUF_SIZE; ++j) {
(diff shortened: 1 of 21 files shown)


fkokosinski:
@googlebot I signed it!

danc86:
Thanks for the PR, this looks like a good start.

Looking at the submodule diff:
https://github.com/antmicro/VexRiscv-verilog/compare/301554ed50130998df7c9b3fa2a900ddbefe2027...181fd82493b9fe0aab905627630fd818bd451ad0
I see that it switches to tcal's submodule for vexriscv, which picks up the new dcache CSRs, and also some other misc changes which are probably not going to affect us:
https://github.com/tcal-x/VexRiscv/compare/21d24eb07fad16fd671b3d4668d302cbaa4549a7...067137d6f02d21646abe8d6968e57322dd8bbe7e

However you also changed the dcache size on all the variants, is there any reason for that?

We also need to figure out a way we can bring these changes in without making a mess of all the submodules (we ideally don't want to switch to an Antmicro fork which points at a tcal fork on an old feature branch). Right now I don't have any better ideas of how to handle it, so I am thinking of just creating a separate long-lived branch here in CFU-Playground where we can accumulate various hacks for measuring SPI flash and then discard it at the end after we have extracted all the useful improvements.

danc86:
I am also wondering if we need similar numbers for icache as well. To be honest I don't understand well enough right now how the two interact or how exactly I can interpret the numbers we are seeing.

common/src/benchmarks.c
```diff
@@ -47,12 +61,25 @@ static void __attribute__((noinline)) do_loads_cached(void) {
     buf[i] = acc;  // inhibit optimization
   }
   int end_time = perf_get_mcycle();
+
   int total_iters = 1024 * 1024;
-  printf("Val:%d  Cycles: %d   Cycles/load: %d\n\n\n", acc,
+  printf("Val:%d  Cycles: %d   Cycles/load: %d\n", acc,
          end_time - start_time, (end_time - start_time) / total_iters);
+
+  get_csr_metrics(&acc_cnt, &refill_cnt, &stall_cnt);
```
danc86, line 69: Can we introduce a macro or helper function to avoid copy-pasting this into all the benchmarks?

common/src/benchmarks.c
```diff
@@ -29,9 +30,22 @@
 #define BUF_SIZE (128 * 1024)  // must be at least 1024
 #endif
 
+#define CSR_ACC_COUNTER     0xcc4
+#define CSR_REFILL_COUNTER  0xcc8
+#define CSR_STALL_COUNTER   0xccc
+
 // Each project should make their own proj_menu.c, which will replace this one.
 
+static void get_csr_metrics(uint32_t *acc, uint32_t *refill, uint32_t *stall) {
```
danc86, line 39: Having these numbers here in the benchmarks is certainly interesting, but the more important place we really need this is for the model evaluation. For example, at the end of `pdti8_classify()` how many cycles did we spend, and how many of these were we stalled on dcache or icache?

danc86:
Just for some extra context, what I want to build is a spreadsheet to keep track of our SPI flash speed improvements, like this:

| Total cycles evaluating pdti8 | CS active time | Cache stall time |  |
| --- | --- | --- | --- |
| 4000000 | 85% | 95% | initial version |
| ... | ... | ... | switched to Litespi |
| ... | ... | ... | use quad output read |
| ... | ... | ... | reduce cs_delay to 2 |
| ... | ... | ... | ... |

Obviously the numbers I wrote above are made up.

But the idea is that with each improvement we should see the total cycle count for evaluating the model goes down, the CS active time as a proportion goes up (representing better utilisation of the flash) and proportion of time stalled goes down (representing less CPU time wasted). The change in numbers will help us to understand if each improvement is really having the effect we think it should have. It will also tell us when we have improved it enough and SPI flash is no longer the bottleneck.

Right now I am not sure how I can determine the proportion of time spent stalled filling caches. Is the "dcache stalls" CSR showing a number of cycles? And thus, if the CPU runs for 1000 cycles and the dcache stall CSR increases by 500 cycles, it means the CPU spent 50% of its time waiting for dcache to fill?

common/src/benchmarks.c
fkokosinski: Sure, I can split it into two macros: one for declaring variables and getting the numbers before a test/benchmark, and one that gets the numbers after running a test/benchmark and prints them.

fkokosinski:
> Looking at the submodule diff:
> [antmicro/VexRiscv-verilog@301554e...181fd82](https://github.com/antmicro/VexRiscv-verilog/compare/301554ed50130998df7c9b3fa2a900ddbefe2027...181fd82493b9fe0aab905627630fd818bd451ad0)
> I see that it switches to tcal's submodule for vexriscv, which picks up the new dcache CSRs, and also some other misc changes which are probably not going to affect us:
> [tcal-x/VexRiscv@21d24eb...067137d](https://github.com/tcal-x/VexRiscv/compare/21d24eb07fad16fd671b3d4668d302cbaa4549a7...067137d6f02d21646abe8d6968e57322dd8bbe7e)

Yes, I must've missed that. I will try to drop the commits that don't affect these CSRs. 

> However you also changed the dcache size on all the variants, is there any reason for that?

The reason for these changes is that in order to measure dcache-related metrics, we need to have a dcache in the first place. One way of ensuring that dcache is present in VexRiscv is setting non-zero data cache size in the command line arguments list.

That’s why I’ve bulk-changed (set to 8192) for each VexRiscv variant—to ensure that every variant will have a dcache present. The chosen size is arbitrary and can be changed to any size in a reasonable range.

> We also need to figure out a way we can bring these changes in without making a mess of all the submodules (we ideally don't want to switch to an Antmicro fork which points at a tcal fork on an old feature branch). Right now I don't have any better ideas of how to handle it, so I am thinking of just creating a separate long-lived branch here in CFU-Playground where we can accumulate various hacks for measuring SPI flash and then discard it at the end after we have extracted all the useful improvements.

I think that creating a new branch changing the target branch of this PR to it would be a good solution for now.

fkokosinski:
> I am also wondering if we need similar numbers for icache as well. To be honest I don't understand well enough right now how the two interact or how exactly I can interpret the numbers we are seeing.

Adding instruction cache metrics would require additional changes to VexRiscv. I’m not sure if it would provide interesting data when it comes to LiteSPI performance.

fkokosinski:
> Right now I am not sure how I can determine the proportion of time spent stalled filling caches. Is the "dcache stalls" CSR showing a number of cycles? And thus, if the CPU runs for 1000 cycles and the dcache stall CSR increases by 500 cycles, it means the CPU spent 50% of its time waiting for dcache to fill?

All numbers from these CSRs are the number of clock cycles when that particular event had happened. For stall, the condition is as follows:
```
io.isStalled := io.mem.cmd.valid & !io.mem.cmd.ready;
[...]
when(cache.io.isStalled) {
    stallCounter := stallCounter + 1
}
```

So, from my understanding, what you are saying is right and it would mean that the CPU was waiting for 50% of the time.

common/src/benchmarks.c
fkokosinski: Good idea, I will add it to all `*_classify` functions.

danc86:
> Adding instruction cache metrics would require additional changes to VexRiscv. I’m not sure if it would provide interesting data when it comes to LiteSPI performance.

Well for HPS at least, the icache is filled from SPI flash same as the dcache. I think any time the code makes a non-local jump it is likely to trigger an icache miss and fill from SPI flash. I'm not sure how often that happens when evaluating the model but it might be a substantial number of times.

The best way to know is to measure it with some CSRs for icache :-)

Unfortunately it looks like we can't just directly re-use tcal's approach in the icache, it seems to be structured quite differently. I don't understand it enough to figure out where we could insert similar counters.

@tcal-x do you have any ideas about counters for icache in Vexriscv? Do you think it's worthwhile?

danc86:
I've created a new branch `hps-faster-flash` in this repo, based on the current `main` branch. If you want to re-target this PR to point at that branch, we can start merging this work. We can go back and tidy up the submodules later if needed.

common/src/benchmarks.c
fkokosinski: Should be done now.

common/src/benchmarks.c
fkokosinski: Should be done now.

fkokosinski:
@danc86 thanks. I've retargeted this PR to the `hps-faster-flash`. I've also dropped the dcache unrelated commits in VexRiscv, regenerated verilog sources and dcache metrics getting/displaying in `*_classify` functions. Let me know if I've missed something.

danc86 (approved):
Thanks @fkokosinski, this looks great!

The dcache stall number I am seeing seems surprisingly small:

```
   474M (   474196859) cycles total
[Dcache accesses] Before: 25761616, After: 54537396, Diff: 28775780
[Dcache refills]  Before: 5066, After: 199248, Diff: 194182
[Dcache stalls]   Before: 183030, After: 190873, Diff: 7843
```

That is on HPS, evaluating pdti8 once, with the current design which is not using Litespi at all. The benchmarks indicates we should expect ~600 cycles per flash load so if it only spent 7843 cycles stalled on dcache, that seems way too low -- that would mean it loaded no more than 13 bytes through dcache for the entire model evaluation.

I think I need to look closer at the dcache counters, maybe the stall counter is not actually counting number of cycles stalled like I thought it would.

Regardless, let's merge this now and we can continue iterating on the measurements.

danc86:
We will need to fix the CI failures before we can merge this. However they seem unrelated to this PR, it looks like a failure building flashrom:

```
BUILDING FLASHROM
Replacing all version templates with 267ca96.
Checking for a C compiler... found.
Target arch is x86
Target OS is Linux
Checking for libusb-1.0 headers... not found.

The following features require libusb-1.0: CONFIG_PICKIT2_SPI CONFIG_STLINKV3_SPI CONFIG_DEDIPROG CONFIG_DEVELOPERBOX_SPI CONFIG_CH341A_SPI CONFIG_DIGILENT_SPI.
Please install libusb-1.0 headers or disable all features
mentioned above by specifying make CONFIG_ENABLE_LIBUSB1_PROGRAMMERS=no
See README for more information.
```

I guess it is probably broken on the main branch now too, I'm not sure what could have changed...

scripts/setup
```diff
@@ -28,9 +28,9 @@ fi
 if ! which verilator >/dev/null; then
     missing+=(verilator libevent-dev libjson-c-dev)
 fi
-if ! (apt list -i | grep libusb-1.0-0-dev) ; then
-    missing+=(libusb-1.0-0-dev libftdi1-dev)
-fi
+#if ! (apt list -i | grep libusb-1.0-0-dev) ; then
```
danc86, line 31: Oh, I missed this change before. This breaks the Github actions, we need libusb headers for compiling flashrom.

fkokosinski: Right, I've also noticed that I've introduced a change in `common/Makefile`, which was meant to be only temporary. Thanks for spotting it, it should be fixed now.

danc86:
The CI is still failing on mnv2 golden tests, can you have a look @fkokosinski ?

fkokosinski:
> The CI is still failing on mnv2 golden tests, can you have a look @fkokosinski ?

@danc86 this issue seems to be caused by the custom CSRs added in VexRiscv, which are not present in emulation. I've added them to Renode platform descriptions in the `mnv2_first` demo. It fixes this issue for me locally. If the workflow that was previously failing passes, I will modify the rest of the platform descriptions.

fkokosinski:
@danc86 the CI passed, so I've modified the rest of the platform description files. Let me know if anything else needs to be done before merging this PR.

danc86:
Looks good now, thanks!

PR #123, opened 2021-06-17
Determine smallest possible tensor arena at compile time.

JosephBushagour:
Previously, our code used one of two constants for either a bigger or smaller tensor arena. This worked well, but  it poses a problem when RAM is limited and the arena size is larger than necessary (for example, when fitting the CFU-Playground on a small board like the Fomu or iCEbreaker). This PR replaces the previous approach with a compile-time determination of the smallest possible tensor arena depending on the models included in the build.


tcal-x (approved):
LGTM

PR #125, opened 2021-06-17
Tail rom

alanvgreen:


alanvgreen:
There's going to be a fair amount of work in this stream, so wanted to get something out for review sooner rather than later.

common/src/tflite.cc
```diff
@@ -171,6 +171,10 @@ void tflite_load_model(const unsigned char* model_data) {
     return;
   }
 
+#ifdef TF_LITE_SHOW_MEMORY_USE
+  interpreter->GetMicroAllocator().PrintAllocations();
+#endif
```
tcal-x, line 179: PrintAllocations() provides the info from arena_used_bytes() and much more, correct?

tcal-x (approved):
LGTM

common/src/tflite.cc
alanvgreen: Yes, that's right.

PR #126, opened 2021-06-18
Fix two SymbiFlow issues; bump Litex/litedram/litespi.

tcal-x:
We get one SymbiFlow fix (force the make to be non-parallel)
with the LiteX bump.  
edit: see https://github.com/enjoy-digital/litex/pull/945

Another fix is to disable jtagbone on Arty
(it uses a bel not yet handled by SymbiFlow).

As a side effect, we get a bunch of fixes on LiteX and liteSPI.


danc86 (commented):
Some good Litex and Litespi fixes in there!

PR #129, opened 2021-06-18
Add support for the iCEbreaker board.

JosephBushagour:
The general SoC workflow doesn't work for the iCEbreaker due to its tiny RAM and required arguments during construction. This PR provides support for the iCEbreaker in the CFU-Playground. These changes will also make it easier to support other small boards.

- The iCEbreaker (and other ice40up5k boards) can't fit the entire software in RAM. To fix this, `.text` and `.rodata` are moved into SPI flash (not overwriting the already present gateware).
- The way `linker.ld` was replaced when `PLATFORM` is `hps` has been generalized to work with all platforms in order to allow board-specific linker files.
- Because serialboot won't work when using SPI flash, a "hook" has been added before trying to connect to the board which allows boards to perform specific functions when loading software (like flashing SPI flash).


soc/board_specific_workflows/ice40up5k.py
```diff
@@ -0,0 +1,67 @@
...
+    derived boards require a bios_flash_offset argument. They also cannot fit
+    entire CFU-Playground into RAM, so they must put read-only data in flash.
+    """
+    def __init__(self,
+                 args: argparse.Namespace,
+                 soc_constructor: Callable[..., litex_soc.LiteXSoC],
+                 builder_constructor: Callable[..., builder.Builder] = None,
+                 warn: bool = True) -> None:
+
+        # Currently only fomu variant is supported.
```
alanvgreen, line 38: This comment was a bit confusing until I read the code and realized it was the variant of the CPU called "fomu", which is generally applicable to Ice40UP5K boards. Maybe remove the comment and let the code + error message communicate the concept?

soc/board_specific_workflows/test_ice40up5k.py
```diff
@@ -0,0 +1,62 @@
...
+import workflow_args
+from litex.soc.integration import builder
+from unittest import mock
+
+
+class Ice40UP5KWorkflowTest(unittest.TestCase):
+    """TestCase for the Ice40UP5KWorkflow."""
+    def setUp(self):
+        self.args = workflow_args.parse_workflow_args(
+            ['--cpu_variant', 'not fomu + cfu...'])
```
alanvgreen, line 27: This seems specific to the test_warning test(). 

I suggest:

self.args = workflow_args.parse_workflow_args([])

Then add the not-fomu cpu_variant arg directly into the dictionary in the test_warning case.

soc/board_specific_workflows/test_ice40up5k.py
```diff
@@ -0,0 +1,62 @@
...
+        return ice40up5k.Ice40UP5KWorkflow(self.args,
+                                           self.soc_constructor,
+                                           mock.create_autospec(
+                                               builder.Builder),
+                                           warn=warn)
+
+    @mock.patch('warnings.warn')
+    def test_warning(self, mock_warner):
+        """Asserts the user is warned when cpu_variant switched."""
+        self.simple_init(warn=True)
```
alanvgreen, line 40: An important thing to test is that cpu_variant is set to the correct value (fomu+cfu) in the resulting workflow object.

soc/board_specific_workflows/test_ice40up5k.py
```diff
@@ -0,0 +1,62 @@
...
+        self.soc_constructor = mock.MagicMock()
+
+    def simple_init(self, warn: bool = False):
+        return ice40up5k.Ice40UP5KWorkflow(self.args,
+                                           self.soc_constructor,
+                                           mock.create_autospec(
+                                               builder.Builder),
+                                           warn=warn)
+
+    @mock.patch('warnings.warn')
```
alanvgreen, line 37: Continuing my "Alan is down on patching" series of  comments: can you use  [warnings.catch_warnings](https://docs.python.org/3/library/warnings.html#available-context-managers) instead of patching it?

soc/board_specific_workflows/ice40up5k.py
```diff
@@ -0,0 +1,67 @@
...
+        if warn and args.cpu_variant != 'fomu+cfu':
+            warnings.warn('Only fomu+cfu variant of Vexriscv supported for' +
+                          f'{type(self).__name__}, not {args.cpu_variant}.\n' +
+                          'Using fomu+cfu.')
+            args.cpu_variant = 'fomu+cfu'
```
alanvgreen, line 43: I think this should be outside the if statement so that it always happens? Otherwise the value of the "warn" parameter is changing the code behavior, not just warning about it.

alanvgreen (commented):
LGTM I'll let tcal give the final approval.

soc/board_specific_workflows/ice40up5k.py
JosephBushagour: Yeah I should have realized that the warning itself acts as a comment. Thanks for this note! Fix pushed.

soc/board_specific_workflows/test_ice40up5k.py
JosephBushagour: You're correct here, I've gone ahead and moved the cpu_variant into only the test_warning case.

soc/board_specific_workflows/test_ice40up5k.py
JosephBushagour: Added :).

soc/board_specific_workflows/test_ice40up5k.py
JosephBushagour: Yes! I had no idea that was something I could do. The patching has been removed.

soc/board_specific_workflows/ice40up5k.py
JosephBushagour: Good catch, fix pushed.

JosephBushagour:
Thanks for the great feedback (as always!). All the requested changes are pushed.

soc/board_specific_workflows/__init__.py
```diff
@@ -13,8 +13,9 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-from general import GeneralSoCWorkflow
 from digilent_arty import DigilentArtySoCWorkflow
+from general import GeneralSoCWorkflow
+from icebreaker import IcebreakerSoCWorkflow
```
tcal-x, line 18: Why was it changed to make "general" not first?
Not something you need to change, but I want to make sure I'm not missing something.

JosephBushagour: Yes I just wanted them to be alphabetical :).

tcal-x (approved):
LGTM

PR #130, opened 2021-06-18
Tail rom 3

alanvgreen:
With the merge of the tail rom 2 PR (#128), loading models can be quite irritating.

This PR allows models to run without outputing the ROMable content.


tcal-x (approved):
LGTM

PR #132, opened 2021-06-19
Make sure TARGET is not passed to the TFLM make.

tcal-x:


alanvgreen (approved):
Thank you!

PR #135, opened 2021-06-23
Fix setup guide with missing steps and more info

marcmerlin:
1) Add info for installing renode
2) fix non working autolink for installing vivado
3) add python virtualenv info
4) add required nmigen
5) better instructions on how to activate the code after make load

### DIFF
--- a/docs/source/setup-guide.rst
+++ b/docs/source/setup-guide.rst
@@ -21,7 +21,7 @@ but is significantly more expensive.
 Step 2: Install Vivado and renode
 ---------------------------------
 
-1. vivaldo
+1. vivado
   
   See https://cfu-playground.readthedocs.io/en/latest/vivado-install.html for a comprehensive guide. 
   Note that the software can take up to 8H to download
@@ -102,7 +102,7 @@ as a useful "minimal" system.
    $ make prog
 
    # Build the RISCV program and load it onto the board
-   # Also starts a terminal ( exit the terminal with CTRL+\ )
+   # Also starts a terminal ( exit the terminal with CTRL+C )
    $ make load
    (...)
    /home/merlin/fpga/CFU-Playground/soc/bin/litex_term --speed 3686400  --kernel /home/merlin/fpga/CFU-Playground/proj/proj_template/build/software.bin /dev/ttyUSB1
@@ -139,4 +139,4 @@ as a useful "minimal" system.
 * Then `1` - Person Detection int8 model
 * Then `g` - Golden tests
 
-If the golden tests pass, then all is well.
+If the golden tests pass, then all is well (remember that you can exit with CTRL+C)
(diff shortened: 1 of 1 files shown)


docs/source/setup-guide.rst
```diff
@@ -18,36 +18,42 @@ but is significantly more expensive.
 .. _`Mouser`: https://au.mouser.com/ProductDetail/Digilent/410-319?qs=%2Fha2pyFaduiP6GD6DfdhNp6rR4rT1KTVOohSnRQ%252BMgra5hr4M7aEiQ%3D%3D 
 
 
-Step 2: Install Vivado
-----------------------
+Step 2: Install Vivado and renode
+---------------------------------
 
-(can take ~8 hours to download software)
+1. vivaldo
```
alanvgreen, line 24: nit: vivado

docs/source/setup-guide.rst
```diff
@@ -80,21 +91,49 @@ as a useful "minimal" system.
...
    $ make clean
 
    # Program the bitstream onto the board. The first run will take several minutes
    # as Vivado synthesizes a bitstream
+   # If this works fine, you will get a chasing LED pattern with the 4 LEDs
    $ make prog
 
    # Build the RISCV program and load it onto the board
-   # Also starts a terminal
+   # Also starts a terminal ( exit the terminal with CTRL+\ )
```
alanvgreen, line 105: I didn't know about Ctrl+\ - I have been mashing ctrl-c. 

Hmm. I just tried Ctrl+\ and found it left my terminal in a bad state. Maybe we should recommend mashing ctrl-C? WDYT?

alanvgreen (approved):
I'm happy for this to go in as-is. Two nits, which I'll leave up to you as to whether to fix or not.

docs/source/setup-guide.rst
marcmerlin: oops, fixed, thanks.

docs/source/setup-guide.rst
marcmerlin: CTRL+C is better actually, I changed it.

marcmerlin:
Thanks for the review @alanvgreen  , PTAL

marcmerlin:
@alanvgreen  because of the perms, you'll need to re-approve and submit, I don't have perms :)

marcmerlin:
Hi @mithro , looks like the approve from @alanvgreen isn't enough to allow me to submit. Are you able to give me an approve that allows submit or submit for me ?

docs/source/setup-guide.rst
tcal-x: The trick is that you need *two* Ctrl-C's to exit litex_term, which may be why Marc originally resorted to Ctrl+\ .    (I'm just catching up on stuff now after returning from vacation.)

Issue #133, opened 2021-06-20
On Icebreaker, TFLM unit test fails, and microbenchmarks hang


tcal-x:
I saw this in `proj/proj_template_v/`, with no models compiled in.  `CONV TEST` fails, but `DEPTHWISE_CONV TEST` passes.

This is likely a generic issue with the limited memory, not specific to Icebreaker.  

I suspect that the stack grows down into the allocated `.data` section, causing corruption/hangs.

* We should run this in Renode to confirm that this is indeed the issue.

* The BUF_SIZE used in the memory benchmarks should be shrunk down, at least for the boards with limited main_ram/sram.

tcal-x:
I tried inserting printfs in the "cached loads" benchmark, after reducing BUF_SIZE to 2kB, and captured the following output before it hang:  `buf: 0x00dee0-0x00fedc   acc: 0x00dedc`

Using `nm` to look at symbol addresses in `.data`, the highest locations I could find were:
```
00003468 b rx_buf
000034e8 b tx_buf
00003568 B __malloc_current_mallinfo
00003590 B errno
00003598 B _ebss
00003598 B _end
```

So it looks like there's plenty of space...maybe it's not actually hanging, or my theory is wrong.

OH -- it didn't hang, it was just *very slow* --- 742 cycles per load when executing out of spiflash!

For this board, the VexRiscv build has NO Icache or Dcache, and the code is located in spiflash, not main_ram or sram.

JosephBushagour:
Good catch on the TFLM unit tests, I was confused because I thought I tested them before submitting the PR. It turns out the unit tests fail only when no models are included and only for the Icebreaker (at least on the PR branch). I'll need to look into why the unit tests are behaving that way, I'm currently not sure why they'd pass when a model is included and only fail for the icebreaker when one isn't. 

As for the benchmarks, a buffer size of 128kB is a bit big for the Icebreaker, I agree we should think about including a configuration that decreases the size a bit.

tcal-x:
Fixed with #136.

PR #136, opened 2021-06-23
Fix and clean up linker script for iCEbreaker.

JosephBushagour:
Previously the linker script for the iCEbreaker board:
- exposed unused symbols
- linked multiple empty input sections
- tried to load the `.bss` segment into spiflash
- had a subtle (but consequential) bug with the linking of the `.data` segment.

The bug with the linking of the `.data` segment allowed static variables to be placed in memory at `0x0`. A valid pointer to this static variable would then be mistaken as a null pointer (because it points to `0x0`), leading to incorrect program behavior. This was the cause of TFLM unit tests failing when no model was included in the binary (#133); the `tflite::testing::kInputShape` static array that is used in many of the convolution unit tests was mistaken as being a null pointer.

This PR fixes all of these issue in the linker script for the iCEbreaker.

### DIFF
--- a/common/_common_soc/1bitsquared_icebreaker/ld/linker.ld
+++ b/common/_common_soc/1bitsquared_icebreaker/ld/linker.ld
@@ -8,57 +8,57 @@ INCLUDE regions.ld
 
 SECTIONS
 {
-	.reserved :
-	{
-		/* Space reserved for gateware, shouldn't step on it. */
-		. += 0x30000;
-	} > spiflash
+    /* .data first to take precedence for explicit code we want in SRAM. */
+    .data : AT (_erodata) /* Loaded in spiflash but runtime copied to SRAM. */
+    {
+        . = ALIGN(4);
+        _fdata = .;  /* Start of .data in SRAM, copied to in crt0. */
 
-	.text :
-	{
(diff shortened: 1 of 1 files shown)


common/_common_soc/1bitsquared_icebreaker/ld/linker.ld
```diff
@@ -8,56 +8,56 @@ INCLUDE regions.ld
...
+    {
+        *(.rodata .rodata*)
+        *(.srodata .srodata*)
+
+        . = ALIGN(4);
+        _erodata = .;  /* Start of .data in spiflash, copied from in crt0. */
+    } > spiflash
 
-	.bss : AT (ADDR(.data) + SIZEOF (.data))
+	.bss :
```
tcal-x, line 49: The indentation for .bss is a little wonky (it was before as well) -- can we fix it here and below?

JosephBushagour: Great catch! Forgot to change over those tabs to spaces when I was modifying the linker file. Fix pushed :).

tcal-x (approved):
LGTM

Issue #115, opened 2021-06-14
Use litex-renode scripts to gernerate Renode configurations


PiotrZierhoffer:
Although CFU-Playground uses any board given as a parameter, Renode tests are specific to some predefined platforms.

My suggested list of steps:

- [ ] Add a dependency for https://github.com/litex-hub/litex-renode
- [ ] Use it in the build flow, so that the repl files are autogenerated.
- [ ] Create a generic RESC file that loads the autogenerated REPL file from a predefined location
- [ ] Update the running flow and instructions, so that it's used
- [ ] Remove old RESC/REPL files
- [ ] Update the testing scripts to use the new flow.
- [ ] Extend the testing infrastructure to test more examples/platforms

PiotrZierhoffer:
Most of these points are done, some that are not (like more examples tested) will be addressed in their respective issues

PR #137, opened 2021-06-24
Add most of LiteX Boards targets to Github Actions workflow

robertszczepanski:
This is a follow-up to #115.

With this PR, Github Actions workflow builds and tests in Renode most of the targets supported by LiteX Boards.
Few targets are excluded due to lack of `main_ram` region or lack of UART serial support (these targets are mentiones in comments). It still tests `proj_template` and `mnv2_first` like it did before. Some targets require additional submodules so these are added in separate commit.

Workflow uses [LiteX-Renode](https://github.com/litex-hub/litex-renode) to generate `litex-vexriscv-tflite.repl` for every target based on its `csr.csv`. This generation is implemented via new Make target `renode-repl` which can also be called locally. Following that I decided to remove old `*.repl` files.

`hps` test is in separate job because it must use custom Renode files. Autogenerated `repl` with `resc` and `robot` for `litex-vexriscv-tflite` don't work in this case.

There is also a fix of `TARGET` flag used in `common/Makefile`. It has the same name as the flag in `proj/proj.mk` but it's supposed to mean something else (filename of final software binary). Without this fix binary was built in a file named exactly as the `TARGET` flag (e.g. `digilent_arty`, there was no extension) instead of `software.elf`. It's also more clear how to change that filename with a flag since user can provide something like `SOFTWARE_ELF=output.elf`.


PiotrZierhoffer:
@tcal-x there are some issues here that i would like to discuss more. I will try to sum them up this evening

robertszczepanski:
Latest commit provides autogenerating all Renode scripts instead of these for `hps` board since these are pretty board specific and can't be easily autogenerated by Litex-Renode. 
All Renode scripts (`.robot`, `.resc`, `.repl`) are now saved in `proj/<proj_name>/build/renode` named as the `TARGET` flag value.
Robot scripts are generated from `proj/<proj_name>/<proj_name>.robot` template where a word `TARGET` is swapped to an actual `TARGET` flag value. It can be easily extended for other projects (right now there's support for `project_template` and `mnv2_first`) by adding `.robot` templates since the generating script (`scripts/generate_renode_scripts.py`) checks if that template exists and generates `.robot` for a target or prints a message that a file is missing.

proj/mnv2_first/renode/hps.resc
```diff
@@ -6,6 +6,6 @@ machine StartGdbServer 10001
 showAnalyzer sysbus.uart
 showAnalyzer sysbus.uart Antmicro.Renode.Analyzers.LoggingUartAnalyzer
 
-sysbus LoadELF $ORIGIN/software.elf
+sysbus LoadELF $ORIGIN/../build/software.elf
```
PiotrZierhoffer, line 9: @tcal-x this is something we need to think about. HPS repl is not well generated by litex-renode. It stays here, but it affects the flow of the CI

PiotrZierhoffer: I think it's an issue on the litex-renode side though, we'll try to learn more

mateusz-holenko:
@tcal-x It's ready to be reviewed, could you please take a look?

tcal-x (commented):
How difficult would it be to have this work interactively?   e.g. I as a user enter proj/proj_template_v/ and type 
```
make TARGET=mist renode
```
I tried executing
```
make renode-scripts TARGET=mist
```
but I got the error
```
ModuleNotFoundError: No module named 'generate-renode-scripts'
```
but perhaps I'm just not following the correct recipe.

robertszczepanski:
@tcal-x Thanks, that was caused by lack of `litex-renode` directory in `PYTHONPATH`. I was testing it with `environment` script so it worked for me but I guess path should be set without it (since `environment` script is not even mentioned in Readme). There was also wrong path in `make renode` recipe to run `.resc` in Renode.

Please, take another look, it should work now as you expected it to.

tcal-x (approved):
LGTM, thanks @robertszczepanski .

PR #138, opened 2021-06-25
add menu items for SPI flash debugging

danc86:
Just a couple of extra menu commands I am using to satisfy my paranoia that we are not reading garbage values from SPI flash, as we work on the faster speeds.

I put them into a new top-level menu since they aren't really benchmarks. Let me know if you would rather have them arranged differently.

### DIFF
--- a/common/src/main.c
+++ b/common/src/main.c
@@ -15,17 +15,19 @@
  */
 
 #include <console.h>
+#include <generated/csr.h>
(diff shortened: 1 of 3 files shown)


danc86:
Oops, I forgot to test this with the non-Litespi configuration. Thanks for catching my mistake, CI!

tcal-x (approved):
LGTM

PR #139, opened 2021-06-25
Enable running benchmarks on CrossLink NX Evaluation board

fkokosinski:
Currently it is not possible to run benchmarks (or any other piece of software) from this repository on CNX EVN board. This PR adds a new platform (based on existing `hps` platform), that enables that by using the on-board SPI flash chip.
Building gateware/software (in project directory, e.g. `proj/proj_template`):
```
UART_SPEED=115200 PLATFORM=cnx_evn TARGET=lattice_crosslink_nx_evn make bitstream software
```
Flashing/uploading using `ecpprog` (in project directory, e.g. `proj/proj_template`):
```
ecpprog -so 2097152 build/software.bin
ecpprog -S ../../soc/build/cnx_evn.proj_template/gateware/cnx_evn_platform.bit
```

### DIFF
--- a/common/_common_soc/lattice_crosslink_nx_evn/ld/linker.ld
+++ b/common/_common_soc/lattice_crosslink_nx_evn/ld/linker.ld
@@ -1,0 +1,58 @@
+INCLUDE output_format.ld
+ENTRY(_start)
+
+__DYNAMIC = 0;
+
+INCLUDE regions.ld
+
+
+SECTIONS
+{
+	.text :
+	{
+		_ftext = .;
+		*(.text.start)
+		*(.text .stub .text.* .gnu.linkonce.t.*)
+		_etext = .;
+	} > spiflash
+
+	.rodata :
+	{
+		. = ALIGN(8);
+		_frodata = .;
+		*(.rodata .rodata.* .gnu.linkonce.r.*)
+		*(.rodata1)
+		*(.srodata .srodata.*)
+		. = ALIGN(8);
+		_erodata = .;
+	} > spiflash
+
+	.data : AT (ADDR(.rodata) + SIZEOF (.rodata))
+	{
+		. = ALIGN(8);
+		_fdata = .;
+		*(.data .data.* .gnu.linkonce.d.*)
+		*(.data1)
+		*(.ramtext .ramtext.*)
+		_gp = ALIGN(16);
+		*(.sdata .sdata.* .gnu.linkonce.s.* .sdata2 .sdata2.*)
+		_edata = ALIGN(16); /* Make sure _edata is >= _gp. */
+	} > sram
+
+	.bss : AT (ADDR(.data) + SIZEOF (.data))
+	{
+		. = ALIGN(16);
+		_fbss = .;
+		*(.dynsbss)
+		*(.sbss .sbss.* .gnu.linkonce.sb.*)
+		*(.scommon)
+		*(.dynbss)
+		*(.bss .bss.* .gnu.linkonce.b.*)
+		*(COMMON)
+		. = ALIGN(8);
+		_ebss = .;
+		_end = .;
+	} > sram
+}
+
+PROVIDE(_fstack = ORIGIN(sram) + LENGTH(sram) - 4);
--- a/common/src/benchmarks.c
+++ b/common/src/benchmarks.c
@@ -23,8 +23,12 @@
 #include "perf.h"
 #include "generated/mem.h"
 
-#ifdef PLATFORM_hps
+#if defined(PLATFORM_hps)
 #define BUF_SIZE (32 * 1024)  // must be at least 1024
+#elif defined(MAIN_RAM_SIZE) && (MAIN_RAM_SIZE < 256 * 1024)
+#define BUF_SIZE (2 * 1024)  // must be at least 1024
+#elif defined(SRAM_SIZE) && (SRAM_SIZE < 256 * 1024)
+#define BUF_SIZE (2 * 1024)  // must be at least 1024
 #else
 #define BUF_SIZE (128 * 1024)  // must be at least 1024
 #endif
--- a/soc/board_specific_workflows/__init__.py
+++ b/soc/board_specific_workflows/__init__.py
@@ -16,6 +16,7 @@
 from digilent_arty import DigilentArtySoCWorkflow
 from general import GeneralSoCWorkflow
 from icebreaker import IcebreakerSoCWorkflow
+from lattice_crosslink_nx_evn import LatticeCrossLinkNXEVNSoCWorkflow
 from workflow_args import parse_workflow_args
 
 
@@ -32,6 +33,7 @@ def workflow_factory(target: str) -> GeneralSoCWorkflow:
     return {
         'digilent_arty': DigilentArtySoCWorkflow,
         '1bitsquared_icebreaker': IcebreakerSoCWorkflow,
+        'lattice_crosslink_nx_evn': LatticeCrossLinkNXEVNSoCWorkflow,
     }.get(target, GeneralSoCWorkflow)
 
 
@@ -39,6 +41,7 @@ __all__ = [
     'DigilentArtySoCWorkflow',
     'GeneralSoCWorkflow',
     'IcebreakerSoCWorkflow',
+    'LatticeCrossLinkNXEVNSoCWorkflow',
     'parse_workflow_args'
     'workflow_factory',
 ]
--- a/soc/board_specific_workflows/lattice_crosslink_nx_evn.py
+++ b/soc/board_specific_workflows/lattice_crosslink_nx_evn.py
@@ -1,0 +1,54 @@
+#!/usr/bin/env python3
+# Copyright 2021 The CFU-Playground Authors
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     https://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import general
+from litex.soc.integration import builder
+from litex.soc.integration import soc as litex_soc
+from litex.soc.integration.soc import SoCRegion
+
+from litespi.modules import MX25L12835F
+from litespi.opcodes import SpiNorFlashOpCodes as Codes
+from litespi.phy.generic import LiteSPIPHY
+from litespi import LiteSPI
+
+KB = 1024
+MB = 1024 * KB
+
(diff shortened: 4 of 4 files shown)


danc86:
Thanks @fkokosinski . Sorry if I missed something, why do we need a totally separate SoC definition for Crosslink NX Eval? Can we use the board definition in litex-boards through `common_soc.py` like all the others instead?

fkokosinski:
> Thanks @fkokosinski . Sorry if I missed something, why do we need a totally separate SoC definition for Crosslink NX Eval? Can we use the board definition in litex-boards through `common_soc.py` like all the others instead?

From what I've gathered, `common_soc.py` takes the `BaseSoC` class defined in LiteX Boards. CrossLink NX EVN's `BaseSoC` doesn't have LiteSPI with SPI flash instantiated, and we need them in order to fit the software part of CFU Playground. I've also analyzed the possibility of using a custom workflow instead (like icebreaker does), but as far as I understand it only allows us to change the BaseSoC's arguments and not modify the SoC itself.

The chip on CrossLink NX EVN provides only 2 LRAMs IIRC, which is way too small to fit the whole software, which makes using the on-board flash chip necessary.

JosephBushagour:
> > Thanks @fkokosinski . Sorry if I missed something, why do we need a totally separate SoC definition for Crosslink NX Eval? Can we use the board definition in litex-boards through `common_soc.py` like all the others instead?
> 
> From what I've gathered, `common_soc.py` takes the `BaseSoC` class defined in LiteX Boards. CrossLink NX EVN's `BaseSoC` doesn't have LiteSPI with SPI flash instantiated, and we need them in order to fit the software part of CFU Playground. I've also analyzed the possibility of using a custom workflow instead (like icebreaker does), but as far as I understand it only allows us to change the BaseSoC's arguments and not modify the SoC itself.
> 
> The chip on CrossLink NX EVN provides only 2 LRAMs IIRC, which is way too small to fit the whole software, which makes using the on-board flash chip necessary.

I don't have too much reference with how fundamental the changes need to be, but If you override the `build_soc` method you can make any changes you want to the SoC before actually building it (alternatively you could override `make_soc` and use your own custom SoC constructor). Perhaps something to be looked into just to avoid growth in the number of similar `.mk` files.

danc86:
Yes, I think we should try to avoid copy-pasting the entire Makefile and SoC/platform Python modules for this new board. I'm worried that we will quickly lose track of all the fixes we will want in both copies (as a first example, my PR#140 would apply to this Makefile too). We may soon want a third similar board as well, the Crosslink Voice and Vision board. I'm not sure if it's similar enough to the eval board that I can reuse the designs as is, or if it might need slight tweaks.

Since Icebreaker has the same problem (no large memories so it must run from SPI flash) I think it would be good to follow the pattern set for that board, if possible. Even if we still need a custom SoC subclass to add in the SPI flash, can we fit it into `common_soc.mk` like the others? That will help to reduce the amount of duplication.

HPS will still be the odd one out but that's okay for now I think.

fkokosinski:
@JosephBushagour thanks for the tip!

@danc86 I've remade these changes so that any necessary modification to the SoC goes through the custom workflow file, without introducing another platform. Please let me know if I've missed something or if anything else needs to be fixed before proceeding with this PR.

danc86:
I like this approach!

I tried building this for Crosslink-NX Eval and Radiant doesn't like the spiflash pin assignments:

```
Running general design DRC...

WARNING - map: Top module port 'spiflash_wp' does not connect to anything.
WARNING - map: Top module port 'spiflash_hold' does not connect to anything.
ERROR - map: Cannot lock port 'spiflash_cs_n' on pad 'E13' which is taken per other location, sysConfig or prohibit attribute/const
raint.
ERROR - map: Cannot lock port 'spiflash_clk' on pad 'E12' which is taken per other location, sysConfig or prohibit attribute/constr
aint.
ERROR - map: Cannot lock port 'spiflash_mosi' on pad 'D13' which is taken per other location, sysConfig or prohibit attribute/const
raint.
ERROR - map: Cannot lock port 'spiflash_miso' on pad 'D15' which is taken per other location, sysConfig or prohibit attribute/const
raint.
INFO - map: Errors found in user's design.  Output files not written. Check map report for more details.
```

Do you know what I might be doing wrong?

danc86:
Oh, I think I see the problem. In `litex_boards/targets/lattice_crosslink_nx_evn.py` there is currently:

```
        platform.add_platform_command("ldc_set_sysconfig {{MASTER_SPI_PORT=SERIAL}}")
```

According to [sysCONFIG Usage Guide for Nexus Platform](https://www.latticesemi.com/view_document?document_id=52790), 7.2. MASTER_SPI_PORT, setting `SERIAL` will prevent user assignment of those SPI pins, it keeps them reserved for sysCONFIG instead.

I commented that line out and now Radiant can synthesise it.

danc86:
I still can't get this to work on my Crosslink-NX Eval Board. I'm not quite sure what is going wrong. Radiant 3.0 synthesises the bitstream successfully, and I tried loading it into configuration RAM:

```
$ ecpprog -S lattice_crosslink_nx_evn.bit
init..
IDCODE: 0x010f1043 does not match :(
ECP5 Status Register: 0x31400000
reset..
ECP5 Status Register: 0x31400e10
programming..
ECP5 Status Register: 0x31400000
Bye.
```

The green DONE LED lights while it's programming, but then the red INITN LED comes on and I get no output on the UART. According to the manual that indicates an error state during configuration.

For completeness I also tried an older Radiant 2.0 version which is probably what we originally used when developing for this board, but the result is the same.

I also tried building with Oxide, using this local hack:

```
--- a/soc/common_soc.mk
+++ b/soc/common_soc.mk
@@ -44,7 +44,7 @@ UART_ARGS=  --uart-baudrate $(UART_SPEED)
 LITEX_ARGS= --output-dir $(OUT_DIR) --csr-csv $(OUT_DIR)/csr.csv $(CFU_ARGS) $(UART_ARGS) $(TARGET_ARGS)

 ifdef USE_SYMBIFLOW
-LITEX_ARGS += --toolchain symbiflow
+LITEX_ARGS += --toolchain oxide --nextpnr-timingstrict
 else
 ifdef USE_VIVADO
 LITEX_ARGS += --toolchain vivado
```

plus the other hack mentioned above for MASTER_SPI_PORT and the other hack for FD1P3BX mentioned in enjoy-digital/litex#907. But it fails timing:

```
Warning: Max frequency for clock 'sdrio_clk$glb_clk': 71.09 MHz (FAIL at 75.00 MHz)
```

Can you give some more detail about how you are testing this on your board @fkokosinski ?

fkokosinski:
> Can you give some more detail about how you are testing this on your board @fkokosinski ?

Sure, I'm currently using the following stack:
- prjoxide 0.1
- Yosys 0.9+3833
- nextpnr-nexus version 167867ff
- ecpprog (for bitstream loading/software flashing)
- CNX EVN board rev b

I didn't use the `--nextpnr-timingstrict` switch before, but after enabling it I'm getting this info back:
```
Info: Max frequency for clock 'sdrio_clk$glb_clk': 60.84 MHz (FAIL at 75.00 MHz)
Info: Max frequency for clock   'por_clk$glb_clk': 495.79 MHz (PASS at 25.00 MHz)
```
but the resulting bitstream works as expected.

We've also tested that the synthesis etc. go through with Radiant 2.0 and result with a working bitstream with at least LedChaser working.

> I get no output on the UART

Have you flashed the software? Currently we don't use LiteX BIOS with this setup, so it's necessary to do so.

fkokosinski:
> Flashing/uploading using `ecpprog` (in project directory, e.g. `proj/proj_template`):
> 
> ```
> ecpprog -so 2097152 build/software.bin
> ecpprog -S ../../soc/build/cnx_evn.proj_template/gateware/cnx_evn_platform.bit
> ```

This have changed, because the offset for software is no longer used, so the right command now to use is:
```
ecpprog -s build/software.bin
```

danc86:
I missed the fact that the software must be at flash offset 0 now, thanks for pointing that out.

I also figured out that I need to build my bitstream for the ES part specifically because its device ID is different. The error in configuration was because the ID in the bitstream doesn't match my device. I can't convince Radiant to recognise the ES part number, but Oxide knows how to fill in the correct device ID if I tweak the platform definition to use the ES part number. So the bitstream configures now.

I still don't see any output on the onboard FTDI second channel UART though.

danc86:
Oh, I just saw the note about needing to populate R15 and R17 resistors in order to get the UART signal. They are not populated on this board, and I don't have a soldering iron at home, hmm... Maybe I can use some spare pins and connect them to a separate UART adapter instead.

fkokosinski:
> Oh, I just saw the note about needing to populate R15 and R17 resistors in order to get the UART signal. They are not populated on this board, and I don't have a soldering iron at home, hmm... Maybe I can use some spare pins and connect them to a separate UART adapter instead.

Ah, I forgot to mention this. I'm using ft232 adapter and I've rerouted uart rx/tx to pmod0 pins 1 (`D10`) and 2 (`D9`) and like this:
```
diff --git a/litex_boards/platforms/lattice_crosslink_nx_evn.py b/litex_boards/platforms/lattice_crosslink_nx_evn.py
index dce1fb3..d8f79f5 100644
--- a/litex_boards/platforms/lattice_crosslink_nx_evn.py
+++ b/litex_boards/platforms/lattice_crosslink_nx_evn.py
@@ -28,8 +28,10 @@ _io = [
     # Requires installation of 0-ohm jumpers R15 and R17 to properly route signals
     # Note that it is R15 and R17, not R16 and R17 as stated in the user guide
     ("serial", 0,
-        Subsignal("rx", Pins("F16"), IOStandard("LVCMOS33")),
-        Subsignal("tx", Pins("F18"), IOStandard("LVCMOS33")),
+        #Subsignal("rx", Pins("F16"), IOStandard("LVCMOS33")),
+        #Subsignal("tx", Pins("F18"), IOStandard("LVCMOS33")),
+        Subsignal("rx", Pins("D10"), IOStandard("LVCMOS33")),
+        Subsignal("tx", Pins("D9"), IOStandard("LVCMOS33")),
     ),

     # Section 7.3 General Purpose LEDs
```

tmichalak:
@danc86 I think we could merge this PR and then consider merging the new main branch to the hps-faster-flash branch in order to land #149. What do you think?

danc86:
Yep, sounds good.

I got a UART adapter hooked up to my Crosslink-NX Eval Board and now I can finally see this working :-)

danc86 (approved):
Thanks @fkokosinski

danc86:
Oh, Github says there is a conflict. Probably on the `BUF_SIZE` definition. Can you please rebase this PR @fkokosinski ?

danc86:
Thanks!

danc86:
I've also merged `main` to the `hps-faster-flash` branch and gave it a quick run-through on the CLNX Eval, so we have these commits on there now too.

I'm a little concerned that we are ignoring the failure to meet timing, I think it may lead to some unexpected behaviour of the SoC. For example when I try the SPI flash non-sequential access (menu 8 -> n) the SoC appears to hang:

```
spiflash> n

Running non-s
```

I didn't see that problem when I was ran that code on HPS (at a clock speed that could meet timing).

So far everything else appears to work though.

PR #143, opened 2021-06-30
alternate demo for users without a PMOD2

marcmerlin:
(my arty came with PMOD3 which is entirely incompatible)


marcmerlin:
Please see resulting page https://cfu-playground--143.org.readthedocs.build/en/143/crash-course/gateware.html#part-1-verilog-programming-with-xilinx-vivado
I was able to do the basic demo without a PMOD2 (of course the switch does not work, but the wave output works, which is what matters)

alanvgreen:
Nice work! Thank you for taking the time to write it up!

In general, putting a speaker directly across a PMOD output like this is questionable advice. However, in this specific case, it's fine because there are 200 ohm current limiting resistors on the JA and JD PMOD I/O pins - see page 1 of the [Arty schematic](https://reference.digilentinc.com/_media/reference/programmable-logic/arty-a7/arty_a7_sch.pdf).

Would you mind adding a note to the effect that the reader should only do this with JA and JD, and that it's not recommended for other I/O ports? If it works in with the text, you might also refer to the top of the https://www.fpga4fun.com/MusicBox.html page where it recommends using a 1K resistor, which would be pretty safe for just about any FPGA.

marcmerlin:
Thanks @alanvgreen . Note that I was using amplified speakers, so the resistor shouldn't be needed there? Either way I added the note you recommended.
PTAL

alanvgreen:
Thank you!

PR #146, opened 2021-07-01
tflite.h: Make C++ only

alanvgreen:
As part of our continuing drive to convert the codebase to C++.

The majority of code changes are due to differences between C and C++ types, particularly in printf() type specifiers.


common/src/models/magic_wand/magic_wand.cc
```diff
@@ -55,16 +55,21 @@ MagicWandResult magic_wand_classify() {
   };
 }
 
+static int interpret_score(uint32_t& score) {
+  return 1000000 * static_cast<int>(*reinterpret_cast<float*>(&score));
```
JosephBushagour, line 59: This will always return 1000000 or 0 because the score is in the range [0.0, 1.0]. If we move the position of the multiplication we can get a more precise print-out.
```suggestion
  return static_cast<int>(*reinterpret_cast<float*>(&score) * 1000000);
```

common/src/models/magic_wand/magic_wand.cc
```diff
@@ -55,16 +55,21 @@ MagicWandResult magic_wand_classify() {
...
-      prefix, (int)(*(float*)&res.wing_score * 1000000),
-      (int)(*(float*)&res.ring_score * 1000000),
-      (int)(*(float*)&res.slope_score * 1000000),
-      (int)(*(float*)&res.negative_score * 1000000));
+      "%s approx-- wing: 0.%06d ring: 0.%06d, slope: 0.%06d, negative: "
+      "0.%06d\n",
+      prefix, interpret_score(res.wing_score) * 1000000,
+      interpret_score(res.ring_score) * 1000000,
+      interpret_score(res.slope_score) * 1000000,
+      interpret_score(res.negative_score) * 1000000);
```
JosephBushagour, line 72: We've doubled-up on the `* 1000000` since it was also added in the `interpret_score` function.
```suggestion
      prefix, interpret_score(res.wing_score),
      interpret_score(res.ring_score),
      interpret_score(res.slope_score),
      interpret_score(res.negative_score));
```

common/src/models/mlcommons_tiny_v01/vww/vww.cc
```diff
@@ -52,7 +52,7 @@ int32_t vww_classify() {
   static void name() {                                                 \
     puts(#name);                                                       \
     tflite_set_input(mlcommons_tiny_v01_vww_dataset[test_index].data); \
-    printf("  result-- score: %d\n", vww_classify());                  \
+    printf("  result-- score: %ld\n", vww_classify());                  \
```
JosephBushagour, line 55: nit: doesn't really matter but consider removing one space to keep the alignment of `\`.
```suggestion
    printf("  result-- score: %ld\n", vww_classify());                 \
```

JosephBushagour (approved):
Looks great Alan! I really appreciate the change away from the silly pointer type-punning in the magic wand model. The two suggestions in `magic_wand.cc` are important for printing out an accurate value in the tests but the other suggestion is just a clang-format thing that isn't necessary.

tcal-x (approved):
LGTM

PR #147, opened 2021-07-02
Capitalization; install virtualenv if necessary; lxterm exit.

tcal-x:
* Make capitalization more consistent 
* `virtualenv` is not installed by default on Ubuntu 20, so include instructions
* Clarify litex_term exit instructions (double Ctrl-C)


marcmerlin (approved):
Cool, thanks for adding the clarifications. It's always hard to remember to add a step that wasn't needed on your (in case my) machine :)

tcal-x:
@marcmerlin thanks, yeah, it's difficult to create instructions that are correct for all platforms.  I remember one time about a year ago,  I tried to follow another set of installation instructions (for an early tflite-micro / zephyr on litex demo, I think), and I was using Ubuntu 19, and it was a nightmare.

So yeah, another reason to start using either make-env (conda) or docker.

PR #149, opened 2021-07-06
Add CS line activity measurement

fkokosinski:
This PR (WIP) introduces a way to measure the on-board SPI flash chip activity (e.g. % time when CS line is asserted).

### DIFF
--- a/common/src/metrics.c
+++ b/common/src/metrics.c
@@ -17,9 +17,48 @@
 #include "metrics.h"
 
 #include <stdint.h>
+#include <generated/csr.h>
 
 void get_csr_metrics(uint32_t *acc, uint32_t *refill, uint32_t *stall) {
   __asm__ volatile ("csrr %0, %1" : "=r"(*acc) : "i"(CSR_ACC_COUNTER));
   __asm__ volatile ("csrr %0, %1" : "=r"(*refill) : "i"(CSR_REFILL_COUNTER));
   __asm__ volatile ("csrr %0, %1" : "=r"(*stall) : "i"(CSR_STALL_COUNTER));
+}
+
+void set_flash_control(uint32_t val) {
+#ifdef LITESPI_CS_COUNTER
+  spi_flash_counter_control_write(val);
+#endif
+}
+
+void spi_flash_counter_prepare(void) {
+#ifdef LITESPI_CS_COUNTER
+  spi_flash_counter_control_reset_write(0);
+  spi_flash_counter_control_enable_write(1);
+#endif
+}
+
+void spi_flash_counter_stop(void) {
+#ifdef LITESPI_CS_COUNTER
+  spi_flash_counter_control_reset_write(0);
+  spi_flash_counter_control_enable_write(0);
+#endif
+}
+
+void spi_flash_counter_restart(void) {
+#ifdef LITESPI_CS_COUNTER
+  spi_flash_counter_control_reset_write(1);
+  spi_flash_counter_control_enable_write(0);
+#endif
+}
+
+void spi_flash_counter_print_ticks(void) {
+#ifdef LITESPI_CS_COUNTER
+  uint32_t cs = spi_flash_counter_cs_ticks_read();
+  uint32_t clk = spi_flash_counter_clk_ticks_read();
+
+  printf("[CS active] %zu out of %zu cycles\n", cs, clk);
+#else
+  printf("[CS active] SPI flash activity counter not present in the SoC\n");
+#endif
 }
--- a/common/src/metrics.h
+++ b/common/src/metrics.h
@@ -16,7 +16,9 @@
 
 #ifndef METRICS_H
 #define METRICS_H
+//#include <generated/mem.h>
 #include <stdint.h>
+#include <stdio.h>
 
 #ifdef __cplusplus
 extern "C" {
@@ -43,6 +45,29 @@ void get_csr_metrics(uint32_t *acc, uint32_t *refill, uint32_t *stall);
            stall_pre, stall_cnt, stall_cnt - stall_pre); \
   } while (0)
 
+void spi_flash_counter_prepare(void);
+void spi_flash_counter_stop(void);
+void spi_flash_counter_restart(void);
+void spi_flash_counter_print_ticks(void);
+
+#define FLASH_SETUP_METRICS \
+  spi_flash_counter_prepare()
+
+#define FLASH_PRINT_METRICS \
+  do { \
+    spi_flash_counter_stop(); \
+    spi_flash_counter_print_ticks(); \
+    spi_flash_counter_restart(); \
+  } while (0)
+
+#define PERF_SETUP_METRICS \
+    DCACHE_SETUP_METRICS; \
+    FLASH_SETUP_METRICS
+
+#define PERF_PRINT_METRICS \
+    DCACHE_PRINT_METRICS; \
+    FLASH_PRINT_METRICS
+  
 #ifdef __cplusplus
 }
 #endif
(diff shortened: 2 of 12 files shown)


tmichalak:
@danc86 shall we create a PR for the `hps-faster-flash` with merges from the latest `main` branch or is it something you wanted to do?

danc86:
I merged `main` to `hps-faster-flash` a short while ago, so you can rebase this PR now.

fkokosinski:
@danc86 I've rebased and cleaned this PR. We're currently displaying the results like this:
```
[CS active] 37913156 out of 92802540 cycles
```
Where `cycles` means the system clock cycles.

soc/board_specific_workflows/lattice_crosslink_nx_evn.py
```diff
@@ -23,9 +23,57 @@
...
+from litex.soc.interconnect.csr import CSRField
+from litex.soc.interconnect.csr import CSRStatus
+from litex.soc.interconnect.csr import CSRStorage
+from litex.soc.interconnect.csr import AutoCSR
+
 KB = 1024
 MB = 1024 * KB
 
+
+class Counter(Module, AutoCSR):
```
danc86, line 36: It might be better to give this a more specific name like SPIFlashCounter so that the corresponding CSR function names in the C code are a bit more obvious.

common/src/metrics.c
```diff
@@ -17,9 +17,30 @@
...
+#endif
+}
+
+void get_flash_ticks(void) {
+#ifdef LITESPI_CS_COUNTER
+  uint32_t* ticks_addr = (uint32_t*)(CSR_COUNTER_COUNTER_ADDR);
+#endif
+
+#ifdef LITESPI_CS_COUNTER
+  printf("[CS active] %llu out of %llu cycles\n",
```
danc86, line 40: Hmm if I'm understanding the meaning of the counters correctly, I think this wording is a bit unclear. We are counting CS active in terms of system clock cycles (which is good, because then we can compare it to the total cycle count for the model evaluation and get a proportion of time it was active for). But the second number is SPI clock cycles not system clock cycles. Not really the same type of cycles.

danc86 (changes requested):

soc/board_specific_workflows/lattice_crosslink_nx_evn.py
```diff
@@ -23,9 +23,57 @@
...
+        self.counter = CSRStatus(description="Counter register",
+            fields=[
+                CSRField("clk_ticks", size=32, description="clk data"),
+                CSRField("cs_ticks", size=32, description="cs data"),
+            ])
+        cnt_clk = Signal(32)
+        cnt_cs = Signal(32)
+
+        # fake clk to syn with sck
+        fake_clk = Signal()
```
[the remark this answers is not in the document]
fkokosinski: Thanks for spotting it, we've removed this signal.

soc/board_specific_workflows/lattice_crosslink_nx_evn.py
fkokosinski: The module's name has been changed.

soc/board_specific_workflows/lattice_crosslink_nx_evn.py
```diff
@@ -23,9 +23,57 @@
...
+    def __init__(self, pads, spiflash):
+        self.control = CSRStorage(description="Counter control register",
+            fields=[
+                CSRField("enable", size=1, description="eneable counter"),
+                CSRField("reset",  size=1, description="reset counter", pulse=1)
+            ])
+
+        self.counter = CSRStatus(description="Counter register",
+            fields=[
+                CSRField("clk_ticks", size=32, description="clk data"),
```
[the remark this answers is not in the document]
fkokosinski: CSRFields descriptions were changed and now should be more understandable.

common/src/metrics.c
fkokosinski: Right, we've changed it so that we're counting the CS# line in terms of system clock cycles *and* we're now counting system clock cycles as well, regardless whether CS# line is asserted or not.

fkokosinski:
@danc86 could you please look at this again? Is there anything that still needs to be done before proceeding with this PR?

danc86:
This is looking good now.

But I tried it and the numbers don't seem to make sense. This is on Crosslink NX Eval board, after running pdti8 model with person input (menu 1->1->3):

```
   309M (   308672952) cycles total
[Dcache accesses] Before: 75069295, After: 104893551, Diff: 29824256
[Dcache refills]  Before: 5084, After: 202406, Diff: 197322
[Dcache stalls]   Before: 172669, After: 192006, Diff: 19337
[CS active] 317864933 out of 268538028 cycles
  result is 226
```

I was expecting the `[CS active]` line total cycles to be almost identical to the cycle count from `mcycle` printed above (maybe a little bit short because there is some overhead in reading out all the values and comparing/printing). But it's actually short by quite a lot.

More importantly, the CS active cycle count is larger than both total cycle counts which should be impossible.

The gateware in the SpiFlashCounter module looks right to me. Is it possible there is a mistake in the C code which is printing the values? Do you get more sensible numbers when you run it?

fkokosinski:
> More importantly, the CS active cycle count is larger than both total cycle counts which should be impossible.

Right, I forgot to change the printing order after changing the order that these values are stored in CSRs. Sorry for that. It should be correct now.

common/src/metrics.h
```diff
@@ -43,6 +45,30 @@ void get_csr_metrics(uint32_t *acc, uint32_t *refill, uint32_t *stall);
...
   } while (0)
 
+#define CLK_OFF 0
+#define CS_OFF 1
+
+void get_flash_ticks(void);
+void set_flash_control(uint32_t val);
+
+#define FLASH_SETUP_METRICS \
+  set_flash_control(0x1)
```
danc86, line 55: Instead of numeric constants 0x1, 0x0, 0x2 can you please use the constants from `generated/csr.h`? I guess this should be

    set_flash_control(1 << CSR_SPI_FLASH_COUNTER_CONTROL_ENABLE_OFFSET);

and similarly below. Or better yet, just use the generated helper functions:

    spi_flash_counter_control_enable_write(1);

It makes it clearer what is being set.

danc86 (commented):
Hmm the numbers still don't seem right to me.

```
   318M (   318332690) cycles total
[Dcache accesses] Before: 4205565323, After: 4236229897, Diff: 30664574
[Dcache refills]  Before: 430294, After: 642200, Diff: 211906
[Dcache stalls]   Before: 233400, After: 260556, Diff: 27156
[CS active] 212941829 out of 268549779 cycles
  result is 226
```

The flash counter total cycles is only 84% of the `mcycle` total cycles. They should be almost identical, shouldn't they? Where are we losing 49 million cycles?

common/src/metrics.h
fkokosinski: Should be done now.

fkokosinski:
> Hmm the numbers still don't seem right to me.

@danc86 Apparently this issue was caused by two factors: too large CSRFields (I didn't check it further, but separating CS and CLK ticks into two CSRs did the trick) and changing the `%llu` format string to `%zu`. It now looks like this for me in the pdti8 demo:
```
339M (   339373354) cycles total
[Dcache accesses] Before: 139590520, After: 194764022, Diff: 55173502
[Dcache refills]  Before: 6159, After: 180866, Diff: 174707
[Dcache stalls]   Before: 374436, After: 399717, Diff: 25281
[CS active] 191062525 out of 348589272 cycles
  result is 226
 ```
So the mcycle register now reports 97% of what SPI flash activity clk ticks CSR reports.

danc86 (approved):
Great. The numbers make much more sense now.

The CS cycle counter accumulates 9-13 million extra cycles but I guess this is just because its counter values are not captured until after the mcycle and dcache stats are collected and printf()'ed through the UART, which must take a fair bit of time.

Anyway, the most important thing is it gives us a good idea of the proportion of time the SPI flash is actively utilised which is what we wanted here.

PR #150, opened 2021-07-07
hps_model: Adds new model

alanvgreen:
Add new model for HPS.


common/src/tflite.cc
```diff
@@ -175,6 +175,7 @@ void tflite_load_model(const unsigned char* model_data,
 }
 
 void tflite_set_input_zeros() {
+void tflite_set_input_zeros(void) {
```
danc86, line 178: :sweat_smile:

common/src/tflite.h
```diff
@@ -29,6 +29,9 @@
 // Sets up TfLite with a given model
 void tflite_load_model(const unsigned char* model_data, unsigned int model_length);
 void tflite_set_input_zeros();
+void tflite_load_model(const unsigned char* model_data,
```
danc86, line 30: Seems like a duplicate definition is being introduced here? It's harmless, since it matches the one on line 30 above, but I guess it would be good to tidy it up.

danc86 (approved):
Very exciting!

common/src/tflite.h
tcal-x: I think it's just a reformat of the same line, due to length.
And line 31 has added the explicit "void".

tcal-x (approved):
LGTM

PR #152, opened 2021-07-07
Automate installing Renode

robertszczepanski:
This is a solution for #4.

`scripts/setup` will download and extract Renode portable to `third_party/renode`. Each time the script is called it checks if Renode exists and checks its version so it will download Renode or throw a warning about wrong version.
All calls of Renode in scripts will now use `third_party/renode/renode` to avoid problems with different versions or with lack of Renode itself.

### DIFF
--- a/proj/proj.mk
+++ b/proj/proj.mk
@@ -104,6 +104,7 @@ COMMON_DIR         := $(CFU_ROOT)/common
 COMMON_FILES	   := $(shell find $(COMMON_DIR) -type f)
 MLCOMMONS_SRC_DIR  := $(CFU_ROOT)/third_party/mlcommons
 SAXON_SRC_DIR      := $(CFU_ROOT)/third_party/SaxonSoc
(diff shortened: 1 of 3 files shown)


tcal-x:
Reminder to tcal: documentation should be updated after this is reviewed and merged (remove instructions for manually installing Renode).

README.md
```diff
@@ -39,6 +39,7 @@ For Arty A7-35T it would be [Vivado](https://www.xilinx.com/support/download.htm
 You always need a toolchain for a target that you choose.
 
 If you want to try things out using [Renode](https://renode.io) simulation, then you don't need either the board or toolchain.
+To install it follow the [Renode guidelines](https://github.com/renode/renode#installation) or run the `setup` script to get Renode Portable.
```
PiotrZierhoffer, line 42: @robertszczepanski  please update this, as we don't really support using installed Renode right now

robertszczepanski: Sure, removed that line.

tcal-x (approved):
LGTM

PR #153, opened 2021-07-07
Tail ROM: finishing touches

alanvgreen:
These changes together have tail rom working and tidied up.

tail_rom is still experimental. If we decide to use this approach in future, there will be further review.

This PR contains some changes to code in common/src, and those change should be reviewed most closely.


common/src/playground_util/pause.h
```diff
@@ -0,0 +1,14 @@
+#ifndef _PLAYGROUND_PAUSE_H
+#define _PLAYGROUND_PAUSE_H
+
+#include <stddef.h>
+
+// Busy wait a short while. Typically used to avoid overflowing UART send
+// buffers
+inline static void pause() {
```
tcal-x, line 8: Have you considered using `busy_wait(unsigned)` or `busy_wait_us(unsigned)` ?   
https://github.com/enjoy-digital/litex/blob/master/litex/soc/software/libbase/system.c#L18-L36
https://github.com/enjoy-digital/litex/blob/master/litex/soc/cores/cpu/vexriscv/system.h#L39-L40

I suppose the question is whether every board will have timer0.   The solution you have in the PR is self-contained and portable.

alanvgreen: IIRC, the issue is that the header files does not work in C++.

tcal-x (approved):
LGTM

PR #154, opened 2021-07-08
Support boards without mcycle write support in tflite.cc

JosephBushagour:
Boards using tiny versions of VexRiscv (like ice40up5k derived boards) don't have the ability to write to the `mcycle` csr. This means incorrect cycle counts are printed in `tflite.cc` because `perf_set_mcycle(0);` is a no-op for these boards. This PR adds support for boards that can't write to `mcycle` by not calling `perf_set_mcycle` in `tflite.cc`.

### DIFF
--- a/common/src/tflite.cc
+++ b/common/src/tflite.cc
@@ -14,6 +14,8 @@
 
 #include "tflite.h"
 
+#include <cstdint>
+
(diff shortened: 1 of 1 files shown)


tcal-x (approved):
LGTM

PR #155, opened 2021-07-09
Add third_party/renode to .gitignore.

JosephBushagour:
Currently `scripts/setup` installs Renode in `third_party/renode`. Git doesn't ignore this directory however. This PR adds `third_party/renode` to our `.gitignore` so that it doesn't show up in "Untracked files:"


mithro (approved):
LGTM!

PR #158, opened 2021-07-14
log TFLite profile in CSV format

danc86:
Restores the previous behaviour, which was accidentally lost in PR #157.


common/src/tflite.cc
```diff
@@ -248,7 +248,7 @@ void tflite_classify() {
   }
   uint32_t end = perf_get_mcycle();
 #ifndef NPROFILE
-  profiler->Log();
+  profiler->LogCsv();
 #endif
   perf_print_all_counters();
```
alanvgreen, line 253: Please also move perf_print_all_counters() inside the #if so performance counters are not printed if profiling has been turned offf.

danc86: Ahh I see. They are always zero if NPROFILE is set because all the code which touches perf counters is compiled out.

danc86:
Hmm seems like CFU-Playground has the setting to deny non-fast-forwards on PR merges. It might be worth changing that setting now that we are regularly merging several PRs per day. It tends to create pointless merge commits (in the best case). I personally like the "rebase if necessary" strategy which matches what Gerrit does.

danc86:
I'll rebase this as soon as the other PR is landed.

tcal-x:
> Hmm seems like CFU-Playground has the setting to deny non-fast-forwards on PR merges. It might be worth changing that setting now that we are regularly merging several PRs per day. It tends to create pointless merge commits (in the best case). I personally like the "rebase if necessary" strategy which matches what Gerrit does.

I'd be happy to change this.

alanvgreen (approved):
LGTM

PR #160, opened 2021-07-14
Add illegal instruction handler for boards using "Fomu" CPU variant.

JosephBushagour:
In an upcoming PR, hardware division will be removed from the "Fomu" CPU variant in order to save valuable LCs. The performance impact of this is minuscule -- for all CFU-Playground models that fit on the ice40up5k boards, no division is present in any of the TFLM ops. That being said, a few division instructions are present in the surrounding code (even with the `-mno-div` flag passed into GCC, division is present in the precompiled libgcc), for these cases an illegal instruction handler is written to emulate `div`, `divu`, `rem`, and `remu` instructions in software.

This PR has no performance impact on boards that aren't using the "Fomu" CPU variant.


mithro:
@JosephBushagour - This seems like it would be good to have in upstream LiteX too?

common/src/crt0-vexriscv.S
```diff
@@ -34,6 +37,78 @@ _start:
 .section .text
 .global  trap_entry
 trap_entry:
+#ifdef CONFIG_CPU_VARIANT_FOMU
```
tcal-x, line 40: Perhaps define a different var that is more descriptive of why we need to use `instruction_handler`, e.g. `CONFIG_DIV_UNIMPLEMENTED`?    We could set it in litex/soc/cores/cpu/vexriscv/core.py when the Fomu variant is chosen, using `soc.add_config("DIV_UNIMPLEMENTED")`.

tcal-x (changes requested):
Overall this looks great!

common/src/base.c
```diff
@@ -25,6 +25,9 @@
 #include <string.h>
 #include <uart.h>
 
+#include "instruction_handler.h"
+#include "riscv.h"
```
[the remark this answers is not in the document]
JosephBushagour: The `csr_read` and `CAUSE_ILLEGAL_INSTRUCTION` macros are used from `riscv.h`. We could replace these with inline assembly and a hard-coded value from the RISC-V spec if we want to.

common/src/crt0-vexriscv.S
JosephBushagour: Yeah that seems like a better approach. I've created a [PR in your fork of LiteX](https://github.com/tcal-x/litex/pull/2), once that's merged I'll update the crt0 in this PR to use `CONFIG_CPU_DIV_UNIMPLEMENTED`.

JosephBushagour: The `#ifdef` has been updated to be conditional on `CONFIG_CPU_DIV_UNIMPLEMENTED`, thanks again for the great suggestion!

tcal-x (approved):

common/src/main.c
```diff
@@ -55,6 +57,9 @@ static struct Menu MENU = {
 #ifdef SPIFLASH_BASE
         MENU_ITEM('8', "SPI Flash Debugging", spiflash_menu),
 #endif
+#ifdef CONFIG_CPU_VARIANT_FOMU
```
tcal-x, line 60: @JosephBushagour  here too, I think it makes sense to use the other config flag.

tcal-x (changes requested):

common/src/main.c
JosephBushagour: Good catch! Fix pushed :).

tcal-x (approved):

PR #161, opened 2021-07-14
Update bit widths of the `cmd_payload_function_id` port in Verilog CFUs.

JosephBushagour:
In `proj/proj_template_v` and `proj/example_cfu_v`, an older CFU module signature was used where `cmd_payload_function_id` was only 3 bits wide. The Verilog works fine from a functional standpoint, but as example projects it's probably best to extend the bit width to 10 bits so both funct7 and funct3 are visible.


tcal-x (approved):
LGTM

PR #162, opened 2021-07-15
tflite.cc improvements

alanvgreen:
Several quality-of-life improvements for tflite.cc

### DIFF
--- a/common/src/tflite.cc
+++ b/common/src/tflite.cc
@@ -202,7 +202,7 @@ void tflite_set_input(const void* data) {
 void tflite_set_input_unsigned(const unsigned char* data) {
   auto input = interpreter->input(0);
   for (size_t i = 0; i < input->bytes; i++) {
-    input->data.int8[i] = (int)data[i] - 127;
+    input->data.int8[i] = static_cast<int>(data[i]) - 128;
   }
   printf("Set %d bytes at %p\n", input->bytes, input->data.int8);
 }
--- a/common/src/tflite.cc
+++ b/common/src/tflite.cc
@@ -41,6 +41,18 @@ void* __dso_handle = &__dso_handle;
 // TfLM global objects
 namespace {
 
+// A profiler that prints a "." for each profile event begun
+class ProgressProfiler : public tflite::MicroProfiler {
+ public:
+  virtual uint32_t BeginEvent(const char* tag) {
+    printf(".");
+    return tflite::MicroProfiler::BeginEvent(tag);
+  }
+
+ private:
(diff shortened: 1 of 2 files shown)


common/src/tflite.cc
```diff
@@ -246,12 +258,13 @@ void tflite_classify() {
   if (kTfLiteOk != interpreter->Invoke()) {
     puts("Invoke failed.");
   }
+  printf("\n");
   uint32_t end = perf_get_mcycle();
 #ifndef NPROFILE
-  profiler->Log();
-#endif
+  profiler->LogCsv();
```
danc86, line 264: I sent this one last night in https://github.com/google/CFU-Playground/pull/158 :-)

danc86 (approved):

tcal-x:
@avg, the test failure can be reproduced if you go to proj/mnv2_first/, run "`make renode`", and then try to run the mnv2_first golden tests.    I see some of the new progress bar (`.....`), but then it seems to hang.

tcal-x:
> @AvG, the test failure can be reproduced if you go to proj/mnv2_first/, run "`make renode`", and then try to run the mnv2_first golden tests. I see some of the new progress bar (`.....`), but then it seems to hang.

@avg, actually I'm not sure what's going on, since the Renode seems to hang at that point (when I run the test as described) even on current main.

tcal-x:
@alanvgreen , I have some of it figured out.   The right command to run it yourself is:
```
make SW_ONLY=1 clean renode
```

Then, with these changes, the mnv2_first golden tests run (they don't hang), but they do fail due to different output values.   Note that we're not using the gateware, we're using the C emulation for the CFU here.   It does pass on current main.

common/src/tflite.cc
```diff
@@ -181,33 +193,33 @@ void tflite_load_model(const unsigned char* model_data,
...
   memcpy(input->data.int8, data, input->bytes);
-  printf("Copied %d bytes at 0x%p\n", input->bytes, input->data.int8);
+  printf("Copied %d bytes at %p\n", input->bytes, input->data.int8);
 }
 
 void tflite_set_input_unsigned(const unsigned char* data) {
   auto input = interpreter->input(0);
   for (size_t i = 0; i < input->bytes; i++) {
-    input->data.int8[i] = (int)data[i] - 127;
+    input->data.int8[i] = static_cast<int>(data[i]) - 128;
```
tcal-x, line 214: @alanvgreen  Do the test golden output values need updating for this change?

alanvgreen:
Pushed a new version which updates the mnv2 golden tests to account for the fix to tflite_set_input_unsigned() function.

PR #163, opened 2021-07-15
hps: add SPI flash activity counter

danc86:
Just a copy-paste of PR #149 for HPS. Like the other performance metrics changes this is going to the `hps-faster-flash` branch only.


danc86:
Ahh this probably breaks the Renode simulation because it is touching a new CSR that isn't simulated...

danc86:
It actually broke the Renode simulation because it changed the CSR layout. The UART address is hardcoded in hps.repl.

Issue #142, opened 2021-06-30
Switch to newlib


alanvgreen:
@danc86 points out that the way CFU-Playground currently builds mixes some elements of GCC's library with LiteX's libbase. In particular, some headers come from LiteX, and some from GCC, while linking has both libbase and GCC's libraries.

In order to avoid subtle bugs, we should find some way to compile the LiteX app without libbase. I believe newlib is the library shipped with riscv64-gcc.

Challenges:
- I/O routines in libbase should be extracted to another codebase. In particular, UART routines.
- Still need access to files in include/generated, both the linker scripts and header files.
- To fully port newlib, need implment "syscall" functions as described in [the Sycalls section of the newlib manual](https://sourceware.org/newlib/libc.html#Syscalls) and William Gatliff's [Porting and Using Newlib in Embedded Systems](https://www.cs.ccu.edu.tw/~pahsiung/courses/esd/resources/newlib.pdf)

mithro:
Might want to look into picolibc - https://github.com/picolibc/picolibc

It would actually be really nice to get LiteX to move away from the custom libbase to something more standard like newlib. With LTO and modern compilers removes much of the reason for libbase to exist.

mithro:
FYI - @mateusz-holenko who has previously worked on getting LTO enabled on LiteX.

PR #165, opened 2021-07-16
report "ticks" to the tflite-micro profiler as cycles / 1024

danc86:
In tflite-micro the profiling code uses int32_t to represent ticks,
presumably because it was intended to be fed from a relatively low
precision timer source and not the raw system clock cycle counter.

The HPS model (unaccelerated) has one layer which takes more than 2**31
cycles and so it was being displayed in the profile as taking negative
time.

Report "ticks" as clock cycles / 1024, instead of raw cycles. The choice
of divisor is arbitrary but 1024 can be implemented in a single shift
instruction and still makes the numbers easy to interpret (humans just
have to think of it as "kilocycles").

Also, for completeness sake, use the full 64-bit cycle counter in the
unlikely event that a single profiling event takes more than 2**32
cycles. This implementation of get_mcycle() is taken from OpenTitan:
https://github.com/lowRISC/opentitan/blob/d9dc048680d89bd630ffae673103509e28017709/sw/device/lib/runtime/ibex.h#L28
based on the recommendation in the RISC-V Instruction Set Manual.

### DIFF
--- a/common/src/tensorflow/lite/micro/micro_time.cc
+++ b/common/src/tensorflow/lite/micro/micro_time.cc
@@ -15,24 +15,36 @@ limitations under the License.
 
 #include "tensorflow/lite/micro/micro_time.h"
 
-static inline unsigned get_mcycle() {
-  unsigned result;
-  asm volatile("csrr %0, mcycle" : "=r"(result));
-  return result;
-}
-
-static inline void set_mcycle(unsigned cyc) {
-  asm volatile("csrw mcycle, %0" ::"r"(cyc));
+// Read the cycle counter.
+//
+// The value of the counter is stored across two 32-bit registers: `mcycle` and
+// `mcycleh`. This function is guaranteed to return a valid 64-bit cycle
+// counter value, even if `mcycle` overflows before reading `mcycleh`.
+//
+// Adapted from: The RISC-V Instruction Set Manual, Volume I: Unprivileged ISA
+// V20191213, pp. 61.
+static inline uint64_t get_mcycle() {
+  uint32_t cycle_low = 0;
+  uint32_t cycle_high = 0;
+  uint32_t cycle_high_2 = 0;
+  asm volatile(
+      "read%=:"
+      "  csrr %0, mcycleh;"     // Read `mcycleh`.
+      "  csrr %1, mcycle;"      // Read `mcycle`.
+      "  csrr %2, mcycleh;"     // Read `mcycleh` again.
+      "  bne  %0, %2, read%=;"  // Try again if `mcycle` overflowed before
+                                // reading `mcycleh`.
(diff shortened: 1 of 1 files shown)


danc86:
I also have a patch to make `perf_get_mcycle()` return the 64-bit value, and convert all usages to uint64_t... but I forgot that we are still stuck using Litex libbase printf() and it doesn't handle `%llu` correctly. Which I already discovered weeks ago, and tried fixing in https://github.com/enjoy-digital/litex/pull/910 but never finished. :-(

A nicer solution might be to switch us over to newlib or picolibc (#142) and then we should get a working `%llu`.

common/src/tensorflow/lite/micro/micro_time.cc
```diff
@@ -15,24 +15,36 @@ limitations under the License.
...
+      : "+r"(cycle_high), "=r"(cycle_low), "+r"(cycle_high_2)
+      :);
+  return (uint64_t) cycle_high << 32 | cycle_low;
 }
 
 namespace tflite {
 
 // Assumes 100 MHz
-int32_t ticks_per_second() { return 100000000; }
+int32_t ticks_per_second() { return 100000000 / 1024; }
```
alanvgreen, line 45: Passing comment, not necessary to make a change in this PR: We ought to be able to use CONFIG_CLOCK_FREQUENCY from generated/soc.h here, right?

alanvgreen (approved):
LGTM. Thank you for addressing this long running (ha! pun!) issue.

Nit: The cycle counter is restarted when an inference begins rather than as each layer begins. So the problem is not that a single layer takes more than 2^31 seconds, rather that the entire evaluation is longer than 2^31 seconds. See https://github.com/tensorflow/tflite-micro/blob/main/tensorflow/lite/micro/micro_profiler.cc#L62

common/src/tensorflow/lite/micro/micro_time.cc
tcal-x: You're right, this is Arty-specific and will be wrong for many other boards.

tcal-x (approved):
LGTM

danc86:
> Nit: The cycle counter is restarted when an inference begins rather than as each layer begins. So the problem is not that a single layer takes more than 2^31 seconds, rather that the entire evaluation is longer than 2^31 seconds. See https://github.com/tensorflow/tflite-micro/blob/main/tensorflow/lite/micro/micro_profiler.cc#L62

The profiler code handles 32-bit wraparound, so if the whole evaluation takes more than 2**31 seconds but each layer takes less than 2**31 seconds everything would be fine. The issue is specifically with one layer taking more than 2**31 which just makes it print as a negative number (because they have used int32_t instead of uint32_t).

alanvgreen:
Yes, you're right.

PR #170, opened 2021-07-18
Update Python CFU interface

alanvgreen:
Brings API into line with other nmigen code by discouraging passing of modules between components.

A simple API for making CFUs with independent instructions is preserved via the simple_cfu() function, but CFUs with interdependent instructions should use a subclass of the CFU class, and instantiate their instructions in the elab_instructions() method. avg_pdti8/cfu.py has an example of a Cfu subclass.

### DIFF
--- a/python/nmigen_cfu/cfu.py
+++ b/python/nmigen_cfu/cfu.py
@@ -192,7 +192,7 @@ class Cfu(SimpleElaboratable):
             self.reset
         ]
 
-    def elab_instructions(self, m: Module) -> dict[int, InstructionBase]:
+    def elab_instructions(self, m):
         """Make instructions this CFU will execute.
 
         Returns:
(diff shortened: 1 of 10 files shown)


alanvgreen:
Repushed to remove newer Python annotation syntax that was causing CI queue to fail.

python/nmigen_cfu/cfu.py
```diff
@@ -200,6 +192,29 @@ def __init__(self, instructions):
             self.reset
         ]
 
+    def elab_instructions(self, m):
+        """Make instructions this CFU will execute.
+
+        Returns:
+          A dictionary with keys in range(8) containing the CFU's instructions.
```
tcal-x, line 199: Is "range(8)" because only funct3 is used for this?

tcal-x: Ah, yes, I see it in the comments above.   I hadn't been aware before.

tcal-x (approved):
LGTM

python/nmigen_cfu/cfu.py
alanvgreen: Thanks, Tim.

PR #171, opened 2021-07-19
hps_accel/blocks.h: New header file.

alanvgreen:
Documents some building blocks for an initial accelerator architecture.

### DIFF
--- a/proj/hps_accel/src/blocks.h
+++ b/proj/hps_accel/src/blocks.h
@@ -22,7 +22,7 @@
 #include <cstdint>
 
 namespace hps_accel {
-// Enocodes 16 signed 8-bit values in a 4x4 matrix
+// Encodes 16 signed 8-bit values in a 4x4 matrix
 struct Matrix4x4 {
   // values is indexed by y. Each 32 bit value holds 4 values, in little endian
   // order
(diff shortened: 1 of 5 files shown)


proj/hps_accel/src/blocks.h
```diff
@@ -0,0 +1,73 @@
...
+// Iterates through filters for each input channel within output channel and
+// restarts from beginning when it reaches the end
+Matrix4x4 GetFilter();
+
+// Loads input values into global input storage.
+//
+// - width is length of row
+// - in_channels are the number of input channels. Must be divisible by 4.
+// - values is address of first word to load. It is the top left pixel.
+void LoadValues(size_t width, size_t in_channels, uint32_t values);
```
danc86, line 73: Should this be `uint32_t *values` (a pointer)?

proj/hps_accel/src/blocks_test.cc
```diff
@@ -0,0 +1,95 @@
...
+      {Matrix4x4::build(-93, 46, -5, -32, -28, 79, 84, 57, 124, 114, 47, -75,
+                        -77, -91, 7, 92),
+       Matrix4x4::build(-86, 39, 120, -60, 10, -7, 2, -88, -23, -69, 106, 36,
+                        -52, -36, -80, -31),
+       93, -19504},
+  };
+
+  size_t cases = 0;
+  size_t failures = 0;
+  for (auto m : multiply_cases) {
```
danc86, line 88: Fancy C++ shenanigans :-)

I better brush up on my C++ ...

proj/hps_accel/src/blocks.h
```diff
@@ -23,7 +23,7 @@
 
 namespace hps_accel {
 // Encodes 16 signed 8-bit values in a 4x4 matrix
-struct Matrix4x4 {
+struct Vector16 {
```
danc86, line 26: The Matrix4x4 name seemed to make more sense to me so I might have to pick your brain about what is happening here

proj/hps_accel/src/blocks.h
alanvgreen: Yes. After a few iterations, I put it back to int8_t const*.

proj/hps_accel/src/blocks_test.cc
alanvgreen: Modern C++ !!

proj/hps_accel/src/blocks.h
alanvgreen: The way we ended up using these is as a Vector of 16 elements rather than as a 4x4 Matrix. This was my "I feel silly" moment yesterday. Fortunately, the vector approach results in much simpler code, at least intially.

alanvgreen:
This is now ready for review.

proj/hps_accel/src/blocks_test.cc
```diff
@@ -52,6 +52,51 @@ bool test_multiply(Vector16 input, Vector16 filter, int32_t input_offset,
   return false;
 }
 
+// 1474 bytes of Shakespeare
```
danc86, line 71: More fun than lorem ipsum I suppose :-)

danc86 (approved):
I think I get it! Tomorrow I will try using these.

PR #172, opened 2021-07-19
update profiler output in step-by-step accelerator guide to account for ticks = cycles / 1024

danc86:
PR #165 changed how we report "ticks" to the tflite-micro profiler
(cycles / 1024, instead of the raw cycle count). Update the example
output to match.

### DIFF
--- a/docs/source/step-by-step.rst
+++ b/docs/source/step-by-step.rst
@@ -101,38 +101,39 @@ this:
 
 ::
 
-    0,DEPTHWISE_CONV_2D,8074244
-    1,DEPTHWISE_CONV_2D,8261757
-    2,CONV_2D,11953686
-    3,DEPTHWISE_CONV_2D,4215107
-    4,CONV_2D,8476340
-    5,DEPTHWISE_CONV_2D,8176370
-    6,CONV_2D,13542375
-    7,DEPTHWISE_CONV_2D,2088669
-    8,CONV_2D,6907128
-    9,DEPTHWISE_CONV_2D,4107783
-    10,CONV_2D,11811746
-    11,DEPTHWISE_CONV_2D,1038518
-    12,CONV_2D,6088804
-    13,DEPTHWISE_CONV_2D,2003871
-    14,CONV_2D,11876651
-    15,DEPTHWISE_CONV_2D,1965732
-    16,CONV_2D,11839822
-    17,DEPTHWISE_CONV_2D,1950569
-    18,CONV_2D,11905968
-    19,DEPTHWISE_CONV_2D,1956535
-    20,CONV_2D,11883747
-    21,DEPTHWISE_CONV_2D,1989537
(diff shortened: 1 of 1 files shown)


docs/source/step-by-step.rst
```diff
@@ -142,14 +143,16 @@ this:
         5    |     0  |     0  |   n/a   |            0
         6    |     0  |     0  |   n/a   |            0
         7    |     0  |     0  |   n/a   |            0
-       183M (   183087673) cycles total
+       146M (   145699910) cycles total
```
JosephBushagour, line 146: I'm not able to replicate this decrease in cycle count with my setup (approx. 20% decrease seems significant), do you know why this could be different for you and I? If I can replicate this, I'll make the numbers consistent throughout the doc.

danc86: Yes I wondered about that. I thought I should make the example output consistent with itself (I thought it would be bad to show a CSV from my faster run with the total cycles from a slower run that wouldn't add up).

Have there been some other recent improvements on the main branch which could have sped things up? Submodule bump in Vexriscv perhaps?

I am using Vivado 2020.2, although I don't see how any timing improvement in PnR could result in fewer cycles here.

Could my Arty board have faster DDR or something like that? Seems implausible. This board says it's rev E.

JosephBushagour: I also have rev E. I'll go ahead and test again directly from main (with a different Arty as well). I vaguely remember you mentioning something about building a more recent toolchain, doubt that would cause such a dramatic change though. Worst case scenario, I can update the numbers with my run so that they can be internally consistent and keep your explanation.

danc86: Hmm I have gcc 10.2, not 8.x as the instructions in the docs recommend.

danc86: I double-checked the timing, built `proj_template` from main branch commit e9f6c26cf8151a4cd568c569abfef4ae15b207d2, no local modifications:

```
   146M (   145807043) cycles total
```

danc86: Ah ha. I tried it again using gcc 8.3 from the prebuilt toolchain mentioned in the docs.

```
   183M (   183328857) cycles total
```

I'm surprised the difference is so drastic. Maybe this means we should update to point at a newer toolchain. I see that Sifive have published prebuilt binaries for gcc 10.2 now.

danc86:
I've amended the PR to show numbers from gcc 8.3, which matches the previous ones.

docs/source/step-by-step.rst
JosephBushagour: Wow, that performance difference is no joke. Definitely something to move near the top of the to-do list, thanks so much for figuring this out :).

JosephBushagour (approved):

PR #173, opened 2021-07-19
Add Workflow for the Fomu.

JosephBushagour:
This PR adds an explicit workflow for the Fomu board.

The majority of this process is similar to PR #129 which adds an explicit workflow to the iCEbreaker, the main differences being the bios isn't loaded on the Fomu and the Fomu requires a `TTY=` flag to be set on the command line.


tcal-x (approved):
LGTM

PR #176, opened 2021-07-20
tflite.cc: Remove unused code

alanvgreen:
Removes commented-out OpResolver code. Since we are committed to using
the AllOpResolver, we do not need to add individual ops.


JosephBushagour (approved):
LGTM

PR #181, opened 2021-07-21
hps_accel/cfu.py: Initial CFU

alanvgreen:
Initial CFU skeleton 

- implements a Ping instructions
- shares some constants with C++ code

Contains some changes to proj.mk that @tcal-x may want to review.


tcal-x (approved):
LGTM

I tested on the Arty-35T board.

PR #183, opened 2021-07-22
Improve error when we can't find target board in LiteX-Boards

JosephBushagour:
This PR changes the error message displayed in `soc/common_soc.py` when we cannot find the target board in LiteX-Boards. The error message now lists out all of the supported boards so it's easier for a user to correct their mistake.

### DIFF
--- a/soc/common_soc.py
+++ b/soc/common_soc.py
@@ -15,17 +15,20 @@
 
 import argparse
 import importlib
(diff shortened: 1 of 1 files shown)


tcal-x (approved):
LGTM

Tested with `make TARGET=x prog` and `make TARGET=x renode`.

PR #185, opened 2021-07-23
New Stream API

alanvgreen:
These are just the basics. It could be improved with a little typechecking, and won't be truly useful until we also have stream wrappers for FIFO and memories, and some of the utility classes of the kind LiteX provides, such as BinaryActor.

Would a appreciate any comments.

I have a further PR coming which will build on this.


tcal-x (approved):
LGTM

PR #186, opened 2021-07-23
hps_accel: use accelerated multiply_accumulate() in convolution

danc86:
@alanvgreen do you think it's okay to limit this to padding=valid for now and ignore padding=same? I see there was a discussion where we settled on padding=valid as a temporary solution, but I guess we will need padding=same support sooner or later?

Also worth noting that this causes the model evaluation to fail because the CFU ops are not fully implemented. If you switch it to software CFU implementation (not turned on by default currently) the golden tests pass but this is a severe speed regression. Should we fix that before merging this PR?

### DIFF
--- a/proj/hps_accel/src/tensorflow/lite/kernels/internal/reference/integer_ops/conv_accel.cc
+++ b/proj/hps_accel/src/tensorflow/lite/kernels/internal/reference/integer_ops/conv_accel.cc
@@ -35,8 +35,9 @@ void ConvPerChannel4x4(
   const int dilation_width_factor = 1;
   TFLITE_DCHECK_EQ(params.dilation_height_factor, 1);
   const int dilation_height_factor = 1;
-  const int pad_width = params.padding_values.width;
-  const int pad_height = params.padding_values.height;
+  TFLITE_DCHECK_EQ(params.padding_type, PaddingType::kValid);
+  TFLITE_DCHECK_EQ(params.padding_values.width, 0);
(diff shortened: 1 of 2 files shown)


alanvgreen:
Yes, padding=valid is good for now. There's no rush to support padding=same.

proj/hps_accel/src/tensorflow/lite/kernels/internal/reference/integer_ops/conv_accel.cc
```diff
@@ -65,51 +69,29 @@ void ConvPerChannel4x4(
...
-                // we have seen so far.
-                // TODO(b/174275578): Add a check to make sure the
-                // accumulator depth is smaller than 2^16.
-                acc += filter_val * (input_val + input_offset);
-              }
-            }
+          for (int i = 0; i < filter_height * filter_width * input_depth / 16; ++i) {
+            Vector16 input = hps_accel::GetInput();
+            Vector16 filter = hps_accel::GetFilter();
+            acc += multiply_accumulate(input, filter, input_offset);
```
alanvgreen, line 94: Oh, that came out neatly.

alanvgreen (approved):

danc86:
Amended this PR to load the filter only once (as suggested by Alan), and to simplify the DCHECKs for input bounds checking (since I wasn't happy with the weird arithmetic I originally wrote, when I looked it over again).

PR #187, opened 2021-07-23
hps_accel/set: New set instruction

alanvgreen:
Adds a set instruction that is used to set configuration registers.

### DIFF
--- a/proj/hps_accel/gateware/hps_cfu.py
+++ b/proj/hps_accel/gateware/hps_cfu.py
@@ -12,10 +12,14 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-from nmigen import *
-from nmigen_cfu import InstructionBase, Cfu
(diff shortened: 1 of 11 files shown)


tcal-x (approved):
LGTM

proj/hps_accel/gateware/hps_cfu.py
```diff
@@ -36,10 +40,40 @@ def elab(self, m):
...
+    """A binary actor that adds one to a stream of integers.
+
+    Used to implement the verification register.
+    """
+
+    def __init__(self):
+        super().__init__(unsigned(32), unsigned(32))
+
+    def transform(self, m, input, output):
+        m.d.comb += output.eq(input + 1)
```
danc86, line 53: It obviously doesn't matter greatly, but the comment for REG_VERIFY in constants.py says that it will negate its input value, not add 1.

PR #192, opened 2021-07-24
Optimisations for hps_accel/software_cfu

alanvgreen:
Reduces HPS inference time from 7m 15s to 1m 54s.

The largest reduction comes from inlining software_cfu functions, allowing the compiler to optimize away two switch statements per simulated cfu op.

This required first allowing software_cfu to be ported from C to C++ so that the C++ implementation could be placed in the software_cfu header file. Since this was something we are working toward anyway, I included it in this PR.


tcal-x (approved):
LGTM

proj/hps_accel/src/software_cfu.h
```diff
@@ -0,0 +1,190 @@
...
+
+inline uint32_t Ping(uint32_t rs1, uint32_t rs2) {
+  uint32_t result = ping_storage;
+  ping_storage = rs1 + rs2;
+  return result;
+}
+
+};  // namespace soft_cfu
+
+inline uint32_t software_cfu(int funct3, int funct7, uint32_t rs1,
```
danc86, line 176: Hmm maybe we should try and fix LTO in GCC, it might let us avoid this kind of "dump everything in the header and `inline` it" optimisation...

Issue #174, opened 2021-07-20
avg_pdti8 does not work


alanvgreen:
The avg_pdti8 project has been broken for an unknown period.

Should it be deleted?

tcal-x:
I noticed also that (1) it also fails with software emulation of the CFU, and (2) model golden tests pass.

So it seems the gateware matches the intended behavior, but the intended behavior may only support the layer parameterizations seen in the pdti8 model, and not be correct for other parameterizations as tested by the TFLite unit tests.

So I suspect it is an easy fix (make sure the specialized kernels are only invoked for the intended parameterizations, otherwise fall back to the general kernel).

alanvgreen:
It looks like there is no software emulation code for this accelerator. I'll add a note to the makefile so people know not to try to enable it.

PR #193, opened 2021-07-24
avg_pdti8/Makefile: note about software cfu

alanvgreen:
Note that no software CFU exists for pdti8, so it should not be expected
to work.

This is a fix for #174.


tcal-x (approved):
LGTM

PR #194, opened 2021-07-26
hps_accel - adds macc hw implementation

alanvgreen:
Adds a hardware implementation for multiply accumulate

### DIFF
--- a/proj/hps_accel/src/blocks.cc
+++ b/proj/hps_accel/src/blocks.cc
@@ -80,16 +80,19 @@ Vector16 Vector16::zeroes() { return Vector16{{0, 0, 0, 0}}; }
 // Performs a 16 x 16 vector multiplication
 int32_t multiply_accumulate(Vector16 input, Vector16 filter,
                             int32_t input_offset) {
-  cfu_set_sw(REG_INPUT_OFFSET, input_offset);
-  cfu_set_sw(REG_MACC_INPUT_0, input.values[0]);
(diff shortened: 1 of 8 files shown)


proj/hps_accel/src/blocks.cc
```diff
@@ -122,10 +125,10 @@ void LoadInput(size_t width, size_t in_channels, const int8_t* values) {
...
 
 Vector16 GetInput() {
-  uint32_t word0 = cfu_get(REG_INPUT_0);
-  uint32_t word1 = cfu_get(REG_INPUT_1);
-  uint32_t word2 = cfu_get(REG_INPUT_2);
-  uint32_t word3 = cfu_get(REG_INPUT_3);
+  uint32_t word0 = cfu_get_sw(REG_INPUT_0);
+  uint32_t word1 = cfu_get_sw(REG_INPUT_1);
+  uint32_t word2 = cfu_get_sw(REG_INPUT_2);
+  uint32_t word3 = cfu_get_sw(REG_INPUT_3);
```
tcal-x, line 131: Is it done this way because we're mixing some HW CFU and some SW CFU emulation, so we can't just use the global switch to pick one or the other?

alanvgreen: Yes, that's right. There is no hw version for these yet, so we pin them to the software_cfu implementation.

tcal-x (approved):
LGTM

danc86 (approved):

PR #196, opened 2021-07-26
Enable QPI mode for Crosslink-NX Evaluation board

fkokosinski:
This PR enables the usage of QPI mode on CNX ENV board. It depends on the changes made in litex-hub/litespi#53 and in enjoy-digital/litex#979. Since we're switching to the flashboot process that LiteX provides, the additional step of using the MiSoC image file writer (`mkmscimg`) is required before flashing the software to insert length and CRC32:
```
python <path_to_litex>/litex/soc/software/mkmscimg.py -o output.bin --little --fbi <path_to_cfu_proj>/build/software.bin
```
PS I've noticed that this repo is not using the upstream version of LiteX and instead is using the fork that @tcal-x provides. What is the reason behind this and are there plans to upstream these changes?

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -3,7 +3,7 @@
 	url = https://github.com/m-labs/migen.git
 [submodule "soc/deps/litex"]
 	path = third_party/python/litex
-	url = https://github.com/tcal-x/litex.git
+	url = https://github.com/antmicro/litex.git
 [submodule "soc/deps/litedram"]
 	path = third_party/python/litedram
 	url = https://github.com/enjoy-digital/litedram.git
--- a/common/_common_soc/lattice_crosslink_nx_evn/ld/linker.ld
+++ b/common/_common_soc/lattice_crosslink_nx_evn/ld/linker.ld
@@ -5,10 +5,12 @@ __DYNAMIC = 0;
 
 INCLUDE regions.ld
 
+FBI_LITEX_OFFSET = 0x8;
+
 
 SECTIONS
 {
-	.text :
+	.text (ORIGIN(spiflash) + FBI_LITEX_OFFSET) :
 	{
 		_ftext = .;
 		*(.text.start)
--- a/soc/board_specific_workflows/lattice_crosslink_nx_evn.py
+++ b/soc/board_specific_workflows/lattice_crosslink_nx_evn.py
@@ -14,11 +14,12 @@
 # limitations under the License.
 
 import general
+from collections import namedtuple
 from litex.soc.integration import builder
 from litex.soc.integration import soc as litex_soc
 from litex.soc.integration.soc import SoCRegion
 
-from litespi.modules import MX25L12835F
+from litespi.modules import MX25L12833F
 from litespi.opcodes import SpiNorFlashOpCodes as Codes
 from litespi.phy.generic import LiteSPIPHY
 from litespi import LiteSPI
@@ -68,27 +69,16 @@ class SpiFlashCounter(Module, AutoCSR):
 
 class LatticeCrossLinkNXEVNSoCWorkflow(general.GeneralSoCWorkflow):
     def make_soc(self, **kwargs) -> litex_soc.LiteXSoC:
-        soc = super().make_soc(
-            integrated_rom_size=0,
-            integrated_rom_init=[],
-            **kwargs
+        soc = super().make_soc(**kwargs)
+        soc.mem_map["spiflash"] = 0x20000000
+        soc.add_spi_flash(
+            module=MX25L12833F(Codes.READ_4_4_4, program_cmd=Codes.PP_1_1_1),
+            mode="4x",
+            clk_freq=soc.sys_clk_freq
         )
+        soc.constants['FLASH_BOOT_ADDRESS'] = soc.mem_map["spiflash"]
 
-        soc.spiflash_region = SoCRegion(0x00000000, 16 * MB, mode="r", cached=True, linker=True)
-        spi_platform = soc.platform.request("spiflash")
-        soc.submodules.spiflash_phy = LiteSPIPHY(
-            spi_platform,
-            MX25L12835F(Codes.READ_1_1_1),
-            default_divisor=1)
-        soc.submodules.spiflash_mmap = LiteSPI(phy=soc.spiflash_phy,
-            clk_freq        = soc.sys_clk_freq,
-            mmap_endianness = soc.cpu.endianness)
-
-        soc.csr.add("spiflash_mmap")
-        soc.csr.add("spiflash_phy")
-        soc.bus.add_slave(name="spiflash", slave=soc.spiflash_mmap.bus, region=soc.spiflash_region)
-        soc.bus.add_region("rom", soc.spiflash_region)
-
+        spi_platform = namedtuple("Pads", ["cs_n"])(cs_n=~soc.spiflash_phy.cs)
         soc.submodules.spi_flash_counter = SpiFlashCounter(spi_platform)
         soc.csr.add("spi_flash_counter")
         soc.constants['LITESPI_CS_COUNTER'] = 1


danc86:
What does "the flashboot process" mean? Does it just mean having Litex BIOS code as an embedded ROM, which then jumps to executing from SPI flash? Thus the change in mapping of SPI flash from 0x00000000 to 0x20000000?

I guess the reason this is needed is so that the code in liblitespi can issue the command to enter QPI mode, right?

On HPS we are trying to avoid having any embedded BIOS and just execute the application directly from SPI flash, because we are so resource constrainted that we cannot really fit even the tiniest program into an embedded ROM. Would it be possible to have the gateware issue the QPI command?

fkokosinski:
> What does "the flashboot process" mean? Does it just mean having Litex BIOS code as an embedded ROM, which then jumps to executing from SPI flash? Thus the change in mapping of SPI flash from 0x00000000 to 0x20000000?

Yes, and we're performing the jump by using "boot processes" provided by LiteX, in this case the flashboot process: https://github.com/enjoy-digital/litex/blob/master/litex/soc/software/bios/boot.c#L581

> I guess the reason this is needed is so that the code in liblitespi can issue the command to enter QPI mode, right?

Yes, we need the code to: 1) enable writing to status/configuration register, 2) enable the usage of all 4 data lines, 3) enter the QPI mode.

> Would it be possible to have the gateware issue the QPI command?

Currently LiteSPI's SPIMaster core doesn't seem to have an option to issue commands without any software control, especially that we need to send 3 of them.

danc86:
Okay. On HPS, steps 1 and 2 are not needed (the Quad Enable bit in the Gigadevice chip is non-volatile and we will set it during flash programming). And we would prefer to avoid having a separate BIOS just to issue the QPI command, because we don't really have the resources for any embedded ROM. But this looks good as a first step. Later on we could look at extending LiteSPI to support issuing the QPI command in gateware, if necessary.

fkokosinski:
@danc86 the changes in LiteX and LiteSPI have been merged. Do we want to bump LiteSPI/LiteX submodules just for this `hps-faster-branch`, or would we rather want to bump them for the `main` branch of CFU Playground?

danc86:
Since these are real speed improvements (not just temporary things to measure the speed) we will want these changes on the `main` branch eventually. But right now we are using tcal's fork of Litex for some Fomu-related stuff I believe.

So for now we can just merge this PR to the `hps-faster-flash` branch, to avoid messing up the Fomu work.

As long as the Litex and Litespi improvements have been merged upstream, we will get them eventually, so that will be fine.

danc86 (approved):

fkokosinski:
@danc86 some changes were necessary because of the code in liblitespi in LiteX (forcing us to use `add_spi_flash`). Please recheck the changes and let me know if anything else needs to be done in this PR.

danc86:
Oops sorry. I see you already pointed out in the PR description that I need to run the mkmscimg tool.

danc86:
I also just realised that I now need to program the flash at offset 2MB not 0.

danc86:
Hmm in spite of all that, the BIOS still fails to load the application:

```
Booting from flash...
Error: Invalid image length 0xcccccccc
No boot medium found
```

0xcc was the pattern you were seeing read back when QPI mode wasn't working correctly, wasn't it? I can't remember what the root cause of that or the solution actually was. Am I still missing some fix somewhere?

fkokosinski:
> 0xcc was the pattern you were seeing read back when QPI mode wasn't working correctly, wasn't it? I can't remember what the root cause of that or the solution actually was. Am I still missing some fix somewhere?

The cause for this behavior was that the SPI flash chip didn't have quad lines enabled and/or didn't have QPI mode enabled. I've tested it a bit and noticed that it sometimes does that, especially when booting for the first time after powering it up (which could explain why I didn't experience this issue when developing).

It seems that LiteX caches some number of bytes of SoCRegion, if it has `cached=True`, which the spiflash region does. It does that before we have quad lines/QPI mode enabled, which results in reading back the `0xcc` byte. Using `reboot` command in LiteX BIOS console seems to get around this problem for me (logs below), but this looks like it could be solved by pre-queueing commands before the MMAP core has the chance to issue its own reads (which is essentially what you mentioned/suggested before, so that we wouldn't be reliant on software control).

> I also just realised that I now need to program the flash at offset 2MB not 0.

The change in the origin address of spiflash region doesn't affect where the data starts, so you shouldn't need to use an offset when flashing the chip.

```
--========== Initialization ============--
Initializing MX25L12833F SPI Flash...
SPI Flash clk configured to 37 MHz
Enabling Quad mode...
Switching to QPI mode...

--============== Boot ==================--
Booting from serial...
Press Q or ESC to abort boot completely.
sL5DdSMmkekro
             Timeout
Booting from flash...
Error: Invalid image length 0xcccccccc
No boot medium found

--============= Console ================--

litex> reboot

[after reboot]

--========== Initialization ============--
Initializing MX25L12833F SPI Flash...
SPI Flash clk configured to 37 MHz
Enabling Quad mode...
Switching to QPI mode...

--============== Boot ==================--
Booting from serial...
Press Q or ESC to abort boot completely.
sL5DdSMmkekro
             Timeout
Booting from flash...
Executing booted program at 0x20000008

--============= Liftoff! ===============--
Hello, World!

CFU Playground
```

danc86:
> The change in the origin address of spiflash region doesn't affect where the data starts, so you shouldn't need to use an offset when flashing the chip.

Ah yes. I was getting mixed up with when we previously had the 2MB offset to separate bitstream from application.

danc86:
Okay, I get the same results. After `reboot` it can read from the flash successfully. I think you're right that it must be trying to service reads on the bus before the BIOS has entered QPI mode.

I think we will need to solve that before we can merge this PR. Having SPI flash only work on the second try is not really good enough unfortunately. :-)

Seems there are two possible solutions: (1) send the Enable QPI command from gateware, or (2) add a CSR which the software can use to indicate the flash has been put into QPI mode.

But option 2 will probably have a similar race condition, right? If the software first issues the QPI enable command, then tries to write a CSR to tell Litespi to use READ_4_4_4, the CPU will probably try to read from the flash in the meantime. We really need a way to tell Litespi to issue the QPI enable command and then immediately use READ_4_4_4 from that point onwards. Which is essentially option 1.

Regarding the Quad Enable bit, on MX25L12833F that bit is also non-volatile, same as GD25L128D. So if we want to just say that a human must manually set that bit once on the board during programming, I am okay with that. It might make the gateware simpler, if it can assume the Quad Enable bit is already set and it just needs to issue the Enable QPI command.

alanvgreen:
It looks like the state transition required for QPI READ_4_4_4 is hard. I'd be very OK with dropping READ_4_4_4 support in favor of READ_1_4_4, at least initially.

READ_4_4_4 takes ~80 clock cycles to read an 8 word cache line.
READ_1_4_4 takes about 86 cycles
READ_1_4_4 with XIP mode takes 78 cycles.

XIP also requires a state transition, but it's easier to implement in gateware, especially if we assume that only read commands will be sent to the SPI flash.

fkokosinski:
@danc86 could you look at this PR again? I've managed to solve the problem (system needing reboot before working properly) by moving `spiflash_freq_init` after the chip has been put in QPI mode (enjoy-digital/litex#1004).

danc86:
Oh, that explains what was filling the cache before QPI mode was enabled!

tmichalak:
@danc86 @alanvgreen I think we can live with the current workaround for enabling QPI for now and focus to speeding up the clock.

danc86 (approved):
Indeed, this looks good to me. Tested on my Crosslink-NX Eval Board. Nice work!

danc86:
Btw I tried using READ_1_4_4 opcode and it doesn't work:

```
--- a/soc/board_specific_workflows/lattice_crosslink_nx_evn.py
+++ b/soc/board_specific_workflows/lattice_crosslink_nx_evn.py
@@ -72,7 +72,7 @@ class LatticeCrossLinkNXEVNSoCWorkflow(general.GeneralSoCWorkflow):
         soc = super().make_soc(**kwargs)
         soc.mem_map["spiflash"] = 0x20000000
         soc.add_spi_flash(
-            module=MX25L12833F(Codes.READ_4_4_4, program_cmd=Codes.PP_1_1_1),
+            module=MX25L12833F(Codes.READ_1_4_4, program_cmd=Codes.PP_1_1_1),
             mode="4x",
             clk_freq=soc.sys_clk_freq
         )
```

That would be a useful stepping stone for us since it doesn't require BIOS cooperation.

I think the first issue is in `spiflash_init()`. It puts the flash into QPI mode whenever it's *capable* of READ_4_4_4, even if a different opcode like READ_1_4_4 is actually *configured*:

```
#ifdef SPIFLASH_MODULE_QPI_CAPABLE
        printf("Switching to QPI mode...\n");
        spiflash_master_write(0x00000035, 1, 1, 0x1);
#endif
```

However, even when I comment that out, it still seems to read garbage data:

```
--========== Initialization ============--
Initializing MX25L12833F SPI Flash...
Enabling Quad mode...
SPI Flash clk configured to 37 MHz

--============== Boot ==================--
Booting from serial...
Press Q or ESC to abort boot completely.
sL5DdSMmkekro
             Timeout
Booting from flash...
Error: Invalid image length 0xcccccc0c
No boot medium found
```

danc86:
Oh, sorry. READ_1_4_4 and READ_4_4_4 are the *same* opcode (0xeb), it just effectively means READ_4_4_4 if you're in QPI mode or READ_1_4_4 if you're not. So what I was saying doesn't quite make sense.

But I think my point still stands. I was expecting the quad I/O mode to work without QPI, but it seems there is something still not quite right in Litespi for this configuration (even if we comment out the QPI enabling write).

PR #198, opened 2021-07-27
Single port memory

alanvgreen:
I'm not completely happy with the implementation - it's too complex, so am continuing to work on refactoring.

Built on #194 (hw macc)

### DIFF
--- a/proj/hps_accel/gateware/sp_mem.py
+++ b/proj/hps_accel/gateware/sp_mem.py
@@ -1,0 +1,170 @@
+# Copyright 2021 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     https://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from migen.fhdl.structure import Mux
+from nmigen import unsigned
+from nmigen.hdl.ast import Shape, Signal
+from nmigen.hdl.dsl import Module
+from nmigen.hdl.mem import Memory
+from nmigen.hdl.rec import DIR_FANOUT, Layout
+from nmigen_cfu.util import SimpleElaboratable, ValueBuffer
+
+from .stream import Sink, Source
+
+
+class MemoryParameters:
+    """Defines a memory.
+
+    Also has convenience functions for defining the Sinks and Sources used in a Memory interface.
+    """
+
+    def __init__(self, *, width, depth):
+        """
+        Args:
+          width: int
+            Width of each memory word
+          depth: int
+            Number of words stored in the memory.
+        """
+        self.width = width
+        self.depth = depth
+
+    def addr_shape(self) -> Shape:
+        """nMigen shape describing the address."""
+        return unsigned((self.depth - 1).bit_length())
+
+    def data_shape(self) -> Shape:
+        """nMigen shape describing a data word."""
+        return unsigned(self.width)
+
+    def write_stream_payload_type(self) -> Layout:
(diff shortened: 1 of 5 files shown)


proj/hps_accel/gateware/stream.py
```diff
@@ -75,7 +75,7 @@ def __init__(self, payload_type):
             ("ready", Shape(), DIR_FANIN),
             ("last", Shape(), DIR_FANOUT),
             ("payload", payload_type, DIR_FANOUT),
-        ])
+        ], src_loc_at=3)
```
danc86, line 78: Such magic, wow

danc86: (I went and looked in nmigen what `src_loc_at` does, and I get it, but its name is not very obvious!)

proj/hps_accel/gateware/sp_mem.py
```diff
@@ -0,0 +1,170 @@
...
+        # to memory.
+        m.d.comb += self.write_sink.ready.eq(1)
+        with m.If(self.write_sink.is_transferring()):
+            m.d.comb += [
+                write_port.en.eq(1),
+                write_port.addr.eq(self.write_sink.payload.addr),
+                write_port.data.eq(self.write_sink.payload.data),
+            ]
+
+        m.d.comb += read_port.en.eq(~write_port.en)
```
danc86, line 133: Let me try to check my understanding based on what we discussed this morning...

Here you are using nmigen Memory, which models a "pseudo dual port memory" (meaning it has one port which can only read, and another port which can only write). You are using it to implement a single port memory, by effectively just turning on only one of either the read port or the write port at any given time, in this statement.

Are some of the synthesis tools smart enough to infer from this that the memory needn't actually be dual ported? But yosys isn't, so it will end up trying to construct a dual-port memory where half the bandwidth is wasted?

On Arty that waste doesn't matter because we have heaps of spare resources. But on Crosslink-NX we will need to avoid nmigen Memory and instead configure the `SP16K` EBRAM primitive. That's important because `SP16K` lets us do up to 36 bit width from each EBRAM, meaning we can do a 128 bit memory using 4 EBRAMs instead of 8.

Am I grokking it?

proj/hps_accel/gateware/stream.py
alanvgreen: src_loc_at tells the machinery how deep in the stack to look for a variable name in order to make a name for the Signal()s. It's only important when dumping VCD files or looking at the generated code.

proj/hps_accel/gateware/sp_mem.py
alanvgreen: Yes, that's all correct.

alanvgreen:
Took a look at refactoring - it's going to be a journey, starting with #200, at least. Since this passes test cases I'm happy to submit it now to unblock you.

proj/hps_accel/gateware/test_sp_mem.py
```diff
@@ -0,0 +1,113 @@
...
+class SinglePortMemoryTest(TestBase):
+    def create_dut(self):
+        params = MemoryParameters(width=16, depth=32)
+        return SinglePortMemory(params)
+
+    def test(self):
+        X = XX = None
+        A, B, C, D, E, F, G, H, I = range(11, 20)
+
+        # Description
```
danc86, line 32: I had to stare at this test case for quite a while but I think it makes sense to me now

danc86 (approved):

proj/hps_accel/gateware/test_sp_mem.py
alanvgreen: Any suggestions about how to format the data table to make it easier to understand would be welcome!

danc86: Ah no, it wasn't a complaint, I like your use of symbols to represent the arbitrary values, that certainly helped make the table easier to read. I think the part that took time is just taking the large number of input signals at each step and making my brain decipher them and form a good model of what the DUT should be doing so that I can satisfy myself that I understand why the expected values are what they are. That's probably just a fact of life with these kinds of gateware unit tests, I can't think of any way to format them to help with that.

PR #199, opened 2021-07-27
Add script to do local install of PrjOxide toolchain.

tcal-x:
Usage:
```
./scripts/install_oxide           # initial install
./scripts/install_oxide update    # pull any changes and rebuild and reinstall
```
Initial installation takes about 10 minutes on my gLinux laptop.

The three required repositories are cloned, not added as submodules, under third_party/.

All generated executables are also located under third_party, installed under third_party/usr/local/bin/.  The `environment` script has been updated to use them.

Once the installation is done, you can use them as follows:
```
make PLATFORM=hps USE_OXIDE=1 prog    # use full yosys/nextpnr-nexus flow
make PLATFORM=hps USE_YOSYS=1 prog    # use yosys + radiant pnr
```

### DIFF
--- a/environment
+++ b/environment
@@ -29,3 +29,11 @@ for i in "${CFU_ROOT}/third_party/python"/* "${CFU_ROOT}/python/"* "${CFU_ROOT}/
 done
 export PYTHONPATH
 
+
+if [ -d "${CFU_ROOT}/third_party/usr/local/bin" ] ; then
+    if [ ! `echo ${PATH} | grep third_party/usr/local/bin` ] ; then
+        PATH="${CFU_ROOT}/third_party/usr/local/bin:${PATH}"
+        echo UPDATED PATH: ${PATH}
+        export PATH
+    fi
+fi
--- a/environment
+++ b/environment
@@ -29,11 +29,10 @@ for i in "${CFU_ROOT}/third_party/python"/* "${CFU_ROOT}/python/"* "${CFU_ROOT}/
 done
 export PYTHONPATH
 
-
-if [ -d "${CFU_ROOT}/third_party/usr/local/bin" ] ; then
-    if [ ! `echo ${PATH} | grep third_party/usr/local/bin` ] ; then
-        PATH="${CFU_ROOT}/third_party/usr/local/bin:${PATH}"
-        echo UPDATED PATH: ${PATH}
-        export PATH
+pathadd() {
+    if [ -d "$1" ] && [[ ":$PATH:" != *":$1:"* ]]; then
+        export PATH="${1}${PATH:+":$PATH"}"
+        echo "Updated PATH: ${PATH}"
     fi
-fi
+}
+pathadd "${CFU_ROOT}/third_party/usr/local/bin"
--- a/scripts/install_oxide
+++ b/scripts/install_oxide
@@ -1,0 +1,81 @@
+#!/bin/bash
+set -e
+
+CFU_ROOT="$(dirname $(dirname $(realpath ${BASH_SOURCE[0]})))"
+
+
+UPDATE=
+if [ $# -gt 1 ] ; then
+  echo "Usage: install_oxide [update]"
+  exit 1
+fi
+if [ $# -gt 0 ] ; then
+  if [ $1 == "update" ] ; then
+    UPDATE=1
+  else
+    echo "Usage: install_oxide [update]"
+    exit 1
+  fi
+fi
+
+
+
+if [ -z "$UPDATE" ] ; then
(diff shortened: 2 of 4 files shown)


mithro:
`USE_YOSYS` seems a strange name. As you use Yosys in both places? `USE_RADIANT`, `USE_OXIDE`?

tcal-x:
> `USE_YOSYS` seems a strange name. As you use Yosys in both places? `USE_RADIANT`, `USE_OXIDE`?

It kind of makes sense if your mindset is that the default is using Radiant.  Then `USE_YOSYS` means you just override synthesis but not PNR.   `USE_OXIDE` overrides the entire Radiant flow to instead use yosys for synthesis and nextpnr-nexus for PnR. 

Perhaps I should change the current Makefile to allow specifying arbitrary flags to pass to the LiteX build; then I wouldn't need to invent new names.

USE_YOSYS is equivalent to adding `--synth-mode=yosys` to the LiteX options.

USE_OXIDE is equivalent to adding `--toolchain=oxide` to the LiteX options.

scripts/install_oxide
```diff
@@ -0,0 +1,81 @@
...
+    echo "Usage: install_oxide [update]"
+    exit 1
+  fi
+fi
+
+
+
+if [ -z "$UPDATE" ] ; then
+
+  echo -n "Install Project Oxide open source toolchain? (ctrl-C to exit)" >&2
```
danc86, line 25: Did you mean to put a `read` command or something after this?

environment
```diff
@@ -29,3 +29,11 @@ for i in "${CFU_ROOT}/third_party/python"/* "${CFU_ROOT}/python/"* "${CFU_ROOT}/
 done
 export PYTHONPATH
 
+
+if [ -d "${CFU_ROOT}/third_party/usr/local/bin" ] ; then
+    if [ ! `echo ${PATH} | grep third_party/usr/local/bin` ] ; then
```
danc86, line 34: Maybe it's not a big deal, but this won't work if $PATH contains spaces, because the backticks operator does word splitting on the substituted value. You can replace the backticks with quoted $() operator to avoid word splitting. As in:

    if [ ! "$(echo $PATH | grep ...

My preferred way of writing this kind of condition to check for a substring, is by testing the exit status of `grep -q`, as in:

    if ! grep -q third_party/usr/local/bin <<<$PATH ; then

danc86 (approved):
Thanks for posting this Tim, it looks fine to me aside from some small nits which I have commented.

I agree that the USE_YOSYS and USE_OXIDE variable names were a bit confusing until you explained to me the three possible flows (Radiant, yosys+Radiant, yosys+nextpnr+prjoxide) but I think it does make sense knowing that. It's not a problem for this PR anyway.

environment
```diff
@@ -29,3 +29,11 @@ for i in "${CFU_ROOT}/third_party/python"/* "${CFU_ROOT}/python/"* "${CFU_ROOT}/
 done
 export PYTHONPATH
 
+
+if [ -d "${CFU_ROOT}/third_party/usr/local/bin" ] ; then
```
alanvgreen, line 33: Equivalent code using bash substring testing (idea via stackoverflow, iirc):

    pathadd() {
        if [ -d "$1" ] && [[ ":$PATH:" != *":$1:"* ]]; then
            export PATH="${1}${PATH:+":$PATH"}"
            echo "Updated PATH: ${PATH}"
        fi
    }
    pathadd "${CFU_ROOT}/third_party/usr/local/bin"

Using a function helps reduce clutter by replacing `${CFU_ROOT}/third_party/usr/local/bin` with a shorter variable name

alanvgreen (approved):
I'm keen to try this!

environment
tcal-x: Thanks Alan, I'll make this change.

environment
tcal-x: Thanks for pointing this out!  I'll use Alan's suggestion below for improving this.

scripts/install_oxide
tcal-x: Correct.   I will just delete this, since it's not changing system state (as in, installing apt packages).

tcal-x:
In the most recent commit, I removed the USE_SYMBIFLOW alias in hps.mk, since by now everyone knows USE_OXIDE.

Now when USE_OXIDE is specified, the `--yosys-abc9` flag is passed to LiteX.  In my experience this always results in better timing.

Unfortunately it is not currently possible to request the use of `-abc9` when using the hybrid Yosys+Radiant flow.

PR #200, opened 2021-07-27
hps_accel/stream.py

alanvgreen:
Various improvements and refactorings related to streams. This PR will follow #198 "Single Port Memory".


proj/hps_accel/gateware/stream/stream.py
```diff
@@ -89,8 +93,70 @@ class Sink(_Endpoint):
...
+    Result:
+      A sequence of statements that connects the two sources.
+    """
+    # Checking to catch simple mistakes
+    assert isinstance(source_in, Source)
+    assert isinstance(source_out, Source)
+    assert source_in.payload_type == source_out.payload_type
+
+    return [
+        source_in.ready.eq(source_out.ready),
```
danc86, line 124: Is this essentially equivalent to `source_in._record.connect(source_out._record)`? Would it be worth using the term "connect" here rather than "glue", and putting it into a method on Source and Sink, by analogy with the existing connect method?

Or is there a reason to prefer the term "glue", maybe some distinction I am not seeing?

proj/hps_accel/gateware/stream/actor.py
```diff
@@ -93,7 +93,11 @@ def __init__(self, input_type, output_type):
         super().__init__(input_type, output_type)
 
     def control(self, m: Module):
-        m.d.comb += flowcontrol_passthrough(self.sink, self.source)
+        m.d.comb += [
```
danc86, line 96: I think this is a nice improvement (BinaryCombinatorialActor type instead of flowcontrol_passthrough function).

proj/hps_accel/gateware/stream/__init__.py
```diff
@@ -46,16 +46,26 @@
...
 https://github.com/nmigen/nmigen/issues/317, and especially the existing design
-of LiteX streams.
+of LiteX streams. It relies heavily on the nmigen Record class for payload
+manipulation.
+
+Much of the power of LiteX streams comes from the ability to programatically
+manipulate stream processing code. Because that metaprogramming code is often
+difficult to write and read, this streams implementation makes a minor
+tradeoffs in favor of simplicity of metaprogramming over "everyday" stream
+use. In particular, the stream has a "payload" attribute to explicitly contain
```
danc86, line 56: Good to know why this difference is intentional

danc86 (approved):

proj/hps_accel/gateware/stream/__init__.py
alanvgreen: Mistake on my part: LiteX does have a separate payload, and then also exposes attributes directly on the Source/Sink via getattr() tricks.

proj/hps_accel/gateware/stream/stream.py
alanvgreen: I have been thinking hard about this over the past few days and trying to work out how best to express it. I think it needs a picture. Let me have a go here.

Sinks and Sources each have two sides, an interface side and an implementation side. 

When using a component, we see the interface side of that component's sinks and sources. Data flows to sinks and from sources. For this we use source.connect(sink).

Conversely, when implementing a component, we see data flowing from sinks and to sources. It's up to the implementer to drive the sink and the source.

Now consider the case of a component that has a Source in its interface and a child component that has a source. Further let us say that the child produces exactly the data required at the interface Source, so we want the child component's source to drive the interface source. 

We can't use source.connect(sink) here, because we have two Sinks. We're connecting the interface side of the child source to the implementation side of the interface source.

This is the case where we use glue_sources(). 

(As an alternative to glue_sources, we could leave it up to the implementer to "glue" the two sinks together, but this seems a fairly common case.)

I think I could maybe work up a parable with USB type a and type b plugs, sockets and cables, but I haven't quite figured out whether that would be more helpful than confusing.

I'm also not super-excited about the name. Some ideas I have been tossing around:

- glue_source_to_source
- glue_source_in_to_source_out
- "drive" instead of "glue"
- "wire" 
- "plumb"

PR #202, opened 2021-07-28
Add CFU definition to `index.rst`.

### LINKED ISSUE #201: CFU not defined on page 1
CFU isn't common enough to not be defined near the top.
in this doc it is defined a bit later on the Overview page: 

The SoC has:
    a VexRiscV soft CPU with a Custom Function Unit (CFU) extension

JosephBushagour:
This PR adds a definition for CFU to `index.rst` in our docs so new readers don't need to guess what it means. The extra whitespace is added as a `.rst` formatting quirk to allow nested unordered lists without making the previous list element bold.

The new modified section is shown below:
image

Resolves #201

### DIFF
--- a/docs/source/index.rst
+++ b/docs/source/index.rst
@@ -10,8 +10,15 @@ The CFU Playground is a collection of software, gateware and hardware configured
 make it easy to:
 
 * Run ML models
+
 * Benchmark and profile performance
+
 * Make incremental improvements
+
+  * In software by modifying source code
+
+  * In gateware with a Custom Function Unit (CFU)
+
 * Measure the results of changes
 
 ML acceleration on microcontroller-class hardware is a new area, and one that, due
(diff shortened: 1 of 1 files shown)


mithro:
FYI - @carlfk

tcal-x:
@JosephBushagour,
How about this right under the page title, *before* "The CFU Playground is a ...":

> "**CFU**" stands for Custom Function Unit:  accelerator hardware that is tightly coupled into the pipeline of a CPU core, to add new custom function instructions that complement the CPU’s standard functions (such as arithmetic/logic operations).

CarlFK:
> right under the page title..

Yes.

Spelling out CFU= "Custom Function Unit" and defining it, and explaining the relevant purpose seems like a good thing to start out the lesson.  (the suggested words are fine.)

> tightly coupled into the pipeline of a CPU core

I did not know that.  makes sense.

tcal-x (approved):
LGTM

PR #206, opened 2021-07-29
kws_micro_accel: Link model, TensorFlow ops in SRAM

JosephBushagour:
This PR adds optimizations to move the model, `CONV_2D`, and `DEPTHWISE_CONV_2D` into SRAM from SPI flash. This utilizes approximately 53 KiB more of the SRAM while providing an approximately 2.78x speed-up in inference (from the previous configuration with everything in SPI flash utilizing quad SPI).

Moving the model to SRAM was previously not possible due to a regression in LiteX-Boards that has been fixed in the recent bump.


tcal-x (approved):
LGTM!

PR #207, opened 2021-07-30
common: add perf_get_mcycle64 function

alanvgreen:
Moves the 64 bit get_mcycle() instruction from micro_time.cc to perf.h
so that it is accessible in other contexts. In particular, use it to
print a more accurate time for long running model evaluations.

### DIFF
--- a/common/src/perf.cc
+++ b/common/src/perf.cc
@@ -20,19 +20,19 @@
 
 unsigned CFU_start_counts[NUM_PERF_COUNTERS];
 
-void perf_print_human(unsigned n) {
+void perf_print_human(uint64_t n) {
   if (n > 9999999) {
-    printf("%6uM", (n + 500000) / 1000000);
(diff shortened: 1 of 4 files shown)


danc86 (approved):
Oops. I had a patch to do exactly this, but I hadn't posted it yet because the %llu format specifier still doesn't work.

We can at least use the millions from `perf_print_human()` though, until we do something about printf.

alanvgreen:
I'll make sure I understand what is going on before submitting this.

danc86:
Specifically, when you use %llu it works correctly for values less than 2^32. For values 2^32 and above it prints the 32 least significant bits. It's gotta be some kind of mess up with the ABI but I couldn't find the problem. See https://github.com/enjoy-digital/litex/pull/910

So we can merge this now but we still have to remember to interpret the raw cycle counts as potentially n+2^32 if applicable.

PR #208, opened 2021-07-30
hps_accel: input store

alanvgreen:
Better handling of src_loc_at for Sinks and Sources. This helps quite a
bit with interpreting vcd files.

### DIFF
--- a/proj/hps_accel/gateware/get.py
+++ b/proj/hps_accel/gateware/get.py
@@ -64,7 +64,13 @@ class GetInstruction(InstructionBase):
     """
 
     # The list of all register IDs that may be fetched
-    REGISTER_IDS = [Constants.REG_MACC_OUT, Constants.REG_VERIFY]
+    REGISTER_IDS = [
+        Constants.REG_INPUT_0,
+        Constants.REG_INPUT_1,
+        Constants.REG_INPUT_2,
+        Constants.REG_INPUT_3,
+        Constants.REG_MACC_OUT,
+        Constants.REG_VERIFY]
 
     def __init__(self):
         super().__init__()
--- a/proj/hps_accel/gateware/get.py
+++ b/proj/hps_accel/gateware/get.py
@@ -13,9 +13,10 @@
 # limitations under the License.
 
 from nmigen import unsigned, Signal
+from nmigen.hdl.ast import Mux
 from nmigen.hdl.dsl import Module
 from nmigen_cfu import InstructionBase
-from util import SimpleElaboratable
+from util import SimpleElaboratable, ValueBuffer
 from .constants import Constants
 from .stream import Sink, Source, glue_sinks
 
@@ -23,7 +24,14 @@ from .stream import Sink, Source, glue_sinks
 class StatusRegister(SimpleElaboratable):
     """A register set by gateware.
 
-    Allows gateware to provide data to a CPU
+    Allows gateware to provide data to a CPU.
+
+    Parameters
+    ----------
+
+    valid_at_reset: bool
+      Whether payload is valid at reset or register ought to wait
+      to transfer a value from its sink.
 
     Attributes
     ----------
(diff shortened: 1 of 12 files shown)


proj/hps_accel/gateware/input_store.py
```diff
@@ -0,0 +1,110 @@
...
+    input: Sink(unsigned(32)), in
+      Words written to input are set into the input store.
+
+    next: Signal(), in
+      Causes the next four words to be displayed at the output.
+
+    num_words: Sink(unsigned(32)), in
+      Number of words in input. When set, resets internal storage.
+
+    output: list of 4 Sink(unsigned(32)), out
```
danc86, line 44: Source not Sink :-)

proj/hps_accel/gateware/hps_cfu.py
```diff
@@ -88,6 +89,18 @@ def connect_macc(self, m, set, macc, get):
...
+    def connect_input_store(self, m, set, get, input_store):
+        m.d.comb += [
+            set.sources[Constants.REG_INPUT_NUM_WORDS].connect(
+                input_store.num_words),
+            set.sources[Constants.REG_SET_INPUT].connect(input_store.input),
+            input_store.output[0].connect(get.sinks[Constants.REG_INPUT_0]),
+            input_store.output[1].connect(get.sinks[Constants.REG_INPUT_1]),
+            input_store.output[2].connect(get.sinks[Constants.REG_INPUT_2]),
+            input_store.output[3].connect(get.sinks[Constants.REG_INPUT_3]),
+            input_store.next.eq(get.read_strobes[Constants.REG_INPUT_3]),
```
danc86: Oooh this was the missing piece I needed to make the flow control from store -> registers work in my design! I forgot about the read_strobes signals.

proj/hps_accel/gateware/get.py
```diff
@@ -32,20 +40,32 @@ class StatusRegister(SimpleElaboratable):
...
       register is always ready to receive new data.
 
-    value: unsigned(32), out
+    invalidate: Signal(), in
+      Causes valid to be deassserted.
+
+    value: Signal(32), out
       The value held by the register. Received from sink.payload.
 
+    valid: Signal(), out
```
danc86, line 49: This is interesting. I ran into this issue last night (lack of flow control between the store and the get registers) but the first solution which came to mind was to make the get registers act more like a proper stream consumer -- when the CPU accesses the register, its sink becomes "ready" and the next value is consumed. The CPU would block until the connected source is "valid".

But that would mean that every value sent to the register must be read exactly once. If the CPU doesn't read it, or wants to read it more than once, everything deadlocks. So it's a fragile design.

Whereas signalling validity of the register value means that the CPU is free to read out the register multiple times, as long as the value is valid, which is more like the traditional behaviour of a CPU register. But then in the doc string you note that each register can only be read a single time?

danc86 (approved):
This is neat, at least 75% better than the design I have been working on 😅

proj/hps_accel/gateware/get.py
alanvgreen: Oops. The design process leaked into the comments. Yes, these were exactly the trade-offs I was making and eventually settled on read-many for now.

It would be possible, without too much effort to make some registers read-once, and I think we will need to do that.

alanvgreen: On re-reading the code and my reply, perhaps I was a little confused. Here is the current state: 

- registers can be set by the gateware many times. They are always ready to receive a new value
- registers can be read by the CPU many times. However, if they are not "valid", then the read will block. This is a different "valid" than the register's sink has.
- valid stays asserted unless:
  - valid_at_reset is false, then valid will be deasserted at reset
  - invalidate is asserted.
- after a value is received from the sink, then valid is asserted.

alanvgreen: Fixed the wrong comment in constants.py.

PR #210, opened 2021-08-04
Update Dockerfile so that image supports "make software".

tcal-x:
I ran into errors when I tried to run `cd proj/proj_template_v && make software` inside a container running the generated image, so I changed `mkdir` to `mkdir -p`, and added `unzip` and `curl` to the installed packages.

My commands to build and execute were:
```
$ docker build -t tims-image .
$ docker run -p 8080:80 -it tims-image  bash
```
and then `cd` to proj/proj_template_v and run `make software`.


tcal-x:
@PiotrZierhoffer fyi

tcal-x:
@PiotrZierhoffer , I noticed that the Dockerfile installs a list of all the needed apt packages itself, instead of relying on the call to ./scripts/setup.   It would be nice to test the completeness of ./scripts/setup here.    I realize we need to make sure 'git' is installed in order to clone the repo in order to get the setup script, but other than that, is there a reason we can't just use ./scripts/setup?  And *then* do `rm -rf /var/lib/apt/lists/*` as the final step?

PR #211, opened 2021-08-04
Bump VexRiscv to fix bug in Fomu variant.

JosephBushagour:
This PR bumps VexRiscv to add a memory and writeback stage (approximately 50 LCs) to the Fomu variant. This fixes a bug that caused the CFU to hang during handshaking.


tcal-x (approved):
LGTM

PR #214, opened 2021-08-04
Add gateware and software CFU with tests to kws_micro_accel.

JosephBushagour:
This PR adds a CFU and a software implementation of that CFU to the `kws_micro_accel` project. The CFU performs either a SIMD MAC or an 8 bit MAC based on the first bit of `funct3`, it can be reset with the first bit of `funct7`. This functionality was chosen after testing other options (including replacing the CFU with a single cycle multiplier) and determining the optionally SIMD MAC was both the most effective choice and easily generalized into all relevant TFLM operations.

Due to resource constraints, the CFU is written in SystemVerilog rather than nMigen. Even with the usage of SystemVerilog, LC utilization on the Fomu is at 99%; for this reason compromises were made with the design and implementation of the CFU.


tcal-x (approved):
LGTM

PR #215, opened 2021-08-05
Remove timer from the LiteX build in Fomu workflow.

JosephBushagour:
This PR removes the `timer0` CSRs and related functionality from the LiteX build in the Fomu workflow. The timer is not required for the CFU-Playground, but is necessary for compiling the bios. Thus, the removal of the timer also necessitates the removal of the LiteX bios; this isn't a problem because we didn't use the LiteX bios in the first place with the Fomu.

Overall this change saves us ≈328 LCs which can be utilized in more important applications (like a larger CFU).

### DIFF
--- a/soc/board_specific_workflows/kosagi_fomu.py
+++ b/soc/board_specific_workflows/kosagi_fomu.py
@@ -91,9 +91,12 @@ class KosagiFomuSoCWorkflow(ice40up5k.Ice40UP5KWorkflow):
         self.bios_flash_offset = 0x60000
 
     def make_soc(self, **kwargs) -> litex_soc.LiteXSoC:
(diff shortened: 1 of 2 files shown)


tcal-x (approved):
LGTM

alanvgreen:
@JosephBushagour Happy to help, but I recall that TIMER0 was your suggestion! I pointed you to the generated/csr.h file as a place where you might find unused peripherals. You looked at all the peripherals referenced there and suggested TIMER0 as a candidate :)

mithro:
@JosephBushagour - I actually think someone had previously tried to upstream support for disabling TIMER0 -- it might be worth looking into again if we have demonstrated another need for it.

PR #216, opened 2021-08-05
Add ability to pass extra arguments into board-specific workflows.

JosephBushagour:
This PR adds an `EXTRA_LITEX_ARGS` command line argument to our build system. These extra arguments can be parsed in board-specific workflows to allow for a more flexible building process; ie. `make prog EXTRA_LITEX_ARGS="--variant a7-100"`.

The Arty workflow has been modified to use these extra args to specify the variant of the Arty board (`a7-35` or `a7-100`).

### DIFF
--- a/soc/board_specific_workflows/digilent_arty.py
+++ b/soc/board_specific_workflows/digilent_arty.py
@@ -13,18 +13,44 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+import argparse
 import general
 from litex.soc.integration import soc as litex_soc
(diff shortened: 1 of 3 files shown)


tcal-x (approved):
LGTM

tcal-x:
This is the command that I have confirmed to work, to target the Arty A7-100T board:
```
make TARGET=digilent_arty EXTRA_LITEX_ARGS="--variant=a7-100" clean prog load
```

Note that the build directory is the same as for the Arty a7-35t variant, so you must do a `clean` if you switch from one to the other.

PR #217, opened 2021-08-09
hps: add platform definition for proto2 board

danc86:


alanvgreen (approved):

danc86:
This works on proto2 hardware for proj_template. Seems like the Renode test is broken by this change though, hmm.

danc86:
Amended to fix the CSR addresses in the Renode definition, which changed when I removed the I2CMaster.

PR #219, opened 2021-08-09
hps: fix IO standard for spiflash4x resource on proto0

danc86:
These pins are in bank 0 which means they can only do LVCMOS18 not
LVCMOS18H. The IO standard was already correctly declared on the
'spiflash' resource, it was just incorrect on 'spiflash4x' (currently
unused).

The proto2 platform definition is already correct, I checked it when
I added it.


tcal-x (approved):
LGTM

Issue #89, opened 2021-05-20
RISC-V toolchain, questions about setup guide.


tcal-x:
https://cfu-playground.readthedocs.io/en/latest/setup-guide.html#step-3-install-riscv-toolchain 

@avg, I can fix stuff, but I'm not sure how it was intended to work:

* Is the `RISCV_DIR` variable actually used anywhere?
* Should we `export` it?
* Don't we still need to add the directory to our PATH in the users `.bashrc`, or should it get added elsewhere (`environment`?   `pyrun`?)

danc86:
The only place I've seen `RISCV_DIR` used is in the Makefile for the TFLM subdirectory of our internal HPS firmware repository (not this one). I suspect that might be why it ended up mentioned here.

In my opinion that Makefile should just assume the necessary toolchain programs are already in the PATH (while allowing them to be overridden with variables like `CC` etc) -- more of a normal convention.

The only reason I can think of why we may have wanted `RISCV_DIR` explicitly set and exported is to make sure the user has built the Sifive toolchain rather than using the one provided by their distro. Debian ships a perfectly working toolchain but it lacks a pre-built newlib so it's not usable for our purposes out of the box. However the error message you get in that situation is pretty clear I think (compiler fails to resolve a bunch of standard library symbols).

PR #222, opened 2021-08-10
docs: fix setup-guide instructions

proppy:
- add $RISCV_DIR to PATH
- use python3 builtin venv package


docs/source/setup-guide.rst
```diff
@@ -69,26 +65,27 @@ Step 5: Setup python virtual environment + setup
 
 Python virtualenv ensures that you have the correct python and modules needed by this code.
 
-If you don't have ``virtualenv``, install it; for example, on a Debian-based Linux such as Ubuntu, use:
+If you don't have the python ``venv`` module, install it; for example, on a Debian-based Linux such as Ubuntu, use:
 
 .. code-block:: bash
 
-   $ apt install python3-virtualenv
```
alanvgreen, line 76: I'm pretty sure that the virtual env is not needed at all any more, so long as all python is run using scripts/pyrun. I'll try this out.

tcal-x: I haven't been using either virtualenv or venv, and it's been working for me.   I think the actual reason why it was added in the docs was to get python3.7 instead of 3.9, in cases where 3.9 is the system default.

In any case, venv is preferable to virtualenv since `-m venv` is in the standard Python library, to my understanding.

tcal-x (approved):
LGTM

docs/source/setup-guide.rst
proppy: > so long as all python is run using scripts/pyrun

It seems that at least the proj `Makefile` is launching its python-based rules with this:
https://github.com/google/CFU-Playground/blob/main/proj/proj.mk#L101

Are we sure that this cascade down to all python invocations of the lower layer of the build process?

JosephBushagour:
Great work! I believe these changes close issue #89.

PR #223, opened 2021-08-10
Modify Fomu workflow and VexRiscv build to save LCs and gain performance.

JosephBushagour:
This PR increases the performance of the Fomu VexRiscv build (adds single-cycle multiply, single-cycle shift, and more icache) and removes the `ctrl` registers (registers for resetting the SoC/CPU) from the Fomu LiteX build. Overall this nets us an extra ≈110 LCs for use in a CFU, while making the VexRiscv build significantly more effective (see https://github.com/tcal-x/pythondata-cpu-vexriscv/pull/3 for more on why the changes were made).

This PR also bumps LiteSPI, where some changes have been made to decrease resource utilization.


JosephBushagour:
Force pushed to get CI to run again, previous failures were for networking reasons not related to the content of the PR.

tcal-x (approved):
LGTM

PR #224, opened 2021-08-10
Update gateware/software CFU and tests in kws_micro_accel.

JosephBushagour:
With the new resources afforded to the Fomu build (see #215 and #223) the kws_micro_accel CFU has been updated with new features.

- It continues to perform (optionally SIMD) multiply-and-accumulate operations but now with a status bit that controls the `InputOffset` value. This allows the CFU to be used during the first layer of convolution (which requires the different offset).
- The post-processing section of convolution and depthwise convolution has been moved mostly into the CFU with the newly added "rounding clamping divide by power of two" and "rounding doubling high 32 bit" operations.
  - We don't have enough DSP tiles in the FPGA to move the entirety of the post-processing into gateware so we utilize the updated single-cycle multiply in our CPU before passing our inputs to the CFU.

The software emulation of the CFU has been updated to conform to these changes and new tests have been added for the new functionality. Finally, the gateware CFU implementation has been split apart into separate SystemVerilog modules.

### DIFF
--- a/proj/kws_micro_accel/cfu.sv
+++ b/proj/kws_micro_accel/cfu.sv
@@ -12,6 +12,10 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+`include "mac.sv"
+`include "rcdbpot.sv"
+`include "rdh.sv"
+
 module Cfu (
   input logic clk,
   input logic reset,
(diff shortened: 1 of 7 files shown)


tcal-x (approved):
LGTM!

PR #226, opened 2021-08-11
script/install_oxide: add missing clang dep

proppy:
Without this compilation will complain that clang is missing when building yosys:
```
BUILDING YOSYS
/bin/sh: line 1: clang: command not found
```


tcal-x (approved):
LGTM

PR #229, opened 2021-08-12
Use the `kws_micro_accel` CFU in software and optimize software ops.

JosephBushagour:
PR #224 introduced new functionality to the `kws_micro_accel` CFU. This PR utilizes these new CFU instructions in optimized `CONV2D` and `DEPTHWISE_CONV2D` kernels.

This decreases the cycles per inference to around 25 million, which, compared to the very initial starting point (before any optimizations) of 1.89 billion cycles, is a bit over a cumulative 75x speedup. Golden tests, TFLM unit tests, and the CFU tests all pass with these new changes.

### DIFF
--- a/proj/kws_micro_accel/src/tensorflow/lite/kernels/internal/reference/integer_ops/kws_conv.h
+++ b/proj/kws_micro_accel/src/tensorflow/lite/kernels/internal/reference/integer_ops/kws_conv.h
@@ -1,0 +1,216 @@
+/* Copyright 2021 The CFU-Playground Authors
+   Copyright 2019 The TensorFlow Authors. All Rights Reserved.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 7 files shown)


proj/kws_micro_accel/src/tensorflow/lite/kernels/internal/reference/integer_ops/kws_conv.h
```diff
@@ -0,0 +1,216 @@
+/* Copyright 2021 The CFU-Playground Authors
```
alanvgreen, line 1: If it's not a huge amount of effort, could you please add this file with the unoptimized op in one commit, then in a second commit, make the optimizing changes. This will help future readers understand the change.

I realize that this may not be easily possible, and if so, don't worry about it.

alanvgreen (approved):

proj/kws_micro_accel/src/tensorflow/lite/kernels/internal/reference/integer_ops/kws_conv.h
JosephBushagour: Unfortunately these changes have been in different scratch directories for so long that I would probably have to redo the PR to build some reasonable ordering. Happily, as you mentioned, the software optimizations aren't too obscure, really just constant folding (some of which the compiler does automatically).

This is great feedback that I didn't consider though (I hadn't thought about how commit history acts as documentation), certainly I'll keep in mind more fine-grained commits for any future work.

tcal-x (approved):

PR #234, opened 2021-08-23
hps_accel: fix off-by-one error with max_index signal width

danc86:
This signal was called max_index (implying its value should range up to
N-1 for N input words) but it was actually being used as num_words (its
value would range up to N for N input words).

The logic using the signal was correct, but the signal's width was
wrongly calculated. If num_words was set equal to the maximum capacity
of the input store, the max_index signal would wrap around to 0 instead
of N.

I noticed this when I was unit-testing the filter store, which reuses the same logic. In the filter store unit tests I was setting the store depth equal to the number of words I was filling. The existing input store tests don't catch this because they always fill the store to much less than its capacity.

Amusingly, this is the exact same mistake I made early on when I was writing the filter store too. :-)

### DIFF
--- a/proj/hps_accel/gateware/input_store.py
+++ b/proj/hps_accel/gateware/input_store.py
@@ -74,7 +74,7 @@ class InputStore(SimpleElaboratable):
             m.d.comb += memory.write_sink.payload.data.eq(self.input.payload),
         with m.If(self.input.is_transferring()):
             m.d.comb += memories[write_index[:2]].write_sink.valid.eq(1),
-            with m.If(write_index == max_index - 1):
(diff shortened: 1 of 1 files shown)


alanvgreen (approved):

alanvgreen (changes requested):
Hmm. This made me realize that one of the InputStore test cases was not doing anything. When I fixed that, I found an interesting bug. Fix is coming.

danc86:
> I'll write up a test case for this, too.

Yes I realised in the shower last night (of course...) that I should have included a test case with this PR. I'm glad you're on top of it though, thanks.

PR #235, opened 2021-08-23
hps_accel: implement filter store registers

danc86:
This is largely a copy-paste from the input store implementation, with constants and memory depth tweaked, as a starting point. We expect the two implementations will start to diverge after this (particularly since the filter store will need to be shrunk to fit inside the fewer EBRAMs on HPS).

The tests in test_filter_store.py are independently implemented, which is why they look quite different to test_input_store.py.

This includes two x'ed out test cases which cover some edge cases around reset behaviour, which I think *should* work, but currently don't. It's probably not too important because we can easily avoid the problematic sequence of operations by just making sure we don't ever do that in software.

### DIFF
--- a/proj/hps_accel/gateware/test_filter_store.py
+++ b/proj/hps_accel/gateware/test_filter_store.py
@@ -1,0 +1,194 @@
+# Copyright 2021 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     https://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+
+from nmigen_cfu import TestBase
+
+from .filter_store import FilterStore
+
+
+class FilterStoreTest(TestBase):
+    def create_dut(self):
+        return FilterStore(depth=12)
+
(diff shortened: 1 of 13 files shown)


danc86:
Amended to turn off writing traces in the new test cases.

proj/hps_accel/src/tensorflow/lite/kernels/internal/reference/integer_ops/conv_accel.cc
```diff
@@ -81,7 +87,12 @@ void ConvPerChannel4x4(
         TFLITE_DCHECK_LE(in_x_origin + filter_width, input_width);
         const int8_t *current_input_data = input_data +
             Offset(input_shape, batch, in_y_origin, in_x_origin, 0);
+
+        TFLITE_DCHECK_LE(
+            input_depth * filter_height * filter_width / 4,
+            MAX_INPUT_WORDS);
```
alanvgreen, line 93: Very nice.

proj/hps_accel/gateware/test_filter_store.py
```diff
@@ -0,0 +1,194 @@
...
+            yield
+            # Now reset and fill again from the start
+            yield from self.reset_and_fill()
+            yield from self.read_out()
+        self.run_sim(process)
+
+    # TODO(dcallagh): fix reset logic so that this case passes.
+    # It currently breaks because the first output values get stuck at their
+    # output sources after the first fill, and remain stuck there during the
+    # second fill and up to the subsequent read.
```
alanvgreen, line 172: I believe the root cause of the bug is in SinglePortMemory. I have a rewrite in progress to simplify that class in order to more easily fix the bug.

alanvgreen (approved):
Looks good.

Love the test cases :)

Will merge now so I can make the changes needed to match the upcoming stream API PR.

PR #237, opened 2021-08-23
hps_accel/test_hps_cfu: turn off vcd file

alanvgreen:
Turns off VCD file generation - this should not have been checked in.


danc86 (approved):
Oh I bet I accidentally left all these turned on in PR #235 as well...

PR #238, opened 2021-08-23
hps_accel/stream: Update stream API

alanvgreen:
Defines a new Stream API
- remove distinction between Source and Sink
- make a StreamDescription class that can be used to document whether
  the Stream endpoint ignores the handshaking rules.


proj/hps_accel/gateware/stream/stream.py
```diff
@@ -14,146 +14,106 @@
...
-
-
-class Sink(_Endpoint):
-    """A stream sink
-
+    @staticmethod
+    def cast(obj, src_loc_at=0, ignores_ready=False, ignores_valid=False):
+        if isinstance(obj, StreamDefinition):
+            return StreamDefinition(
+                paylad_type=obj.payload_type,
```
danc86, line 37: paylad -> payload

danc86 (approved):
Very interesting.

I think the Litex name `streams.Endpoint` might be better than `Stream`. I think of the "stream" as really the abstract movement of data over the signals between two "endpoints" which have been connected.

alanvgreen:
I'm very much up in the air with naming. I somewhat agree with your objection to Stream, but I don't think Endpoint is quite right either - when connected, two Endpoints disappear in gateware, so they're no longer endpoints, they're actually the stream. 

I also want to note that, in some situations, three "Endpoints" are connected in a single Stream `A==B==C`. It doesn't seem right to call B an Endpoint in this situation.

Happy to discuss further.

proj/hps_accel/gateware/stream/stream.py
alanvgreen: Hmm. More test cases needed, clearly.

alanvgreen: The new version of this function is used in the normal path, so this is now tested - at least cursorily tested.

danc86 (approved):
I was wondering about the purpose for `ignores_ready` and `ignores_valid`. I guess the idea is to have the other end raise an exception if it doesn't like being connected to streams which behave that way? But since we don't have any like that, there's no need for it now?

alanvgreen:
> I was wondering about the purpose for `ignores_ready` and `ignores_valid`. I guess the idea is to have the other end raise an exception if it doesn't like being connected to streams which behave that way?

Yes, and also formal verification. I don't think this API was a good direction to take. There's a bit more thinking documented at https://github.com/nmigen/nmigen/issues/317#issuecomment-903202512.

> But since we don't have any like that, there's no need for it now?

I'm not sure the value outweighs the impact on the API. Even if we wanted something like this, there are probably better ways to it.

PR #244, opened 2021-08-26
hps_accel: shrink filter store to 2048 words

danc86:
The first 18 convolution layers of the HPS model can still load all
their filter values in one go, outside the loop. The final 9 larger
layers will instead need to load the filter in batches inside the loop,
to fit within the 2048-word capacity.

This gets the filter store down to consuming 4 EBRAMs on HPS.

### DIFF
--- a/proj/hps_accel/src/tensorflow/lite/kernels/internal/reference/integer_ops/conv_accel.cc
+++ b/proj/hps_accel/src/tensorflow/lite/kernels/internal/reference/integer_ops/conv_accel.cc
@@ -71,10 +71,23 @@ void ConvPerChannel4x4(
   const int output_height = output_shape.Dims(1);
   const int output_width = output_shape.Dims(2);
 
+  // Our filter store is of limited size.
+  // First, figure out how many output channels' worth we can fit.
+  bool filter_load_needed;
   TFLITE_DCHECK_LE(
-      input_depth * filter_height * filter_width * output_depth / 4,
+      input_depth * filter_height * filter_width / 4,
       MAX_FILTER_WORDS);
-  hps_accel::LoadFilter(input_depth, output_depth, filter_data);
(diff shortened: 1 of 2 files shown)


proj/hps_accel/src/tensorflow/lite/kernels/internal/reference/integer_ops/conv_accel.cc
```diff
@@ -94,6 +107,14 @@ void ConvPerChannel4x4(
         hps_accel::LoadInput(input_width, input_depth, current_input_data);
 
         for (int out_channel = 0; out_channel < output_depth; ++out_channel) {
+          if (filter_load_needed &&
```
alanvgreen, line 110: I think this commit should go in *as-is*, because it unblocks other work, but wanted to make a note about an alternate implementation:

Given `N = ceil(output_depth / out_channels_per_filter_load)` and `M = out_x * out_y * input_depth` then:

- this commit requires `M` input loads and `NM` filter loads.
- alternative: if we process all the input once per filter load, we instead get `NM` input loads, but only `N` filter loads. 

This alternative is still the same processing complexity (i.e `O(NM)`), but given that input values come from RAM and filter values come from SPI there is potential for significant difference. We should test performance against the models we plan to run.

mnv2_first implements this alternative, and it is quite complex to efficiently calculate the index into output_data.

alanvgreen (approved):
Thanks!

proj/hps_accel/src/tensorflow/lite/kernels/internal/reference/integer_ops/conv_accel.cc
danc86: Yes I wondered about rearranging things here so that we could further reduce the number of filter loads for the large layers. But as you say, it makes the indexes trickier to compute. I went with this version first as it required essentially no changes to the structure of the loops or the index lookups.

PR #247, opened 2021-08-30
hps_accel: connect stores directly to macc

danc86:
This is my first attempt from last week, which uses an explicitly register write from the CPU to "turn the crank" on each iteration.

### DIFF
--- a/proj/hps_accel/src/software_cfu.h
+++ b/proj/hps_accel/src/software_cfu.h
@@ -64,16 +64,17 @@ class Storage {
     len = num_words;
     index = 0;
   }
-  void Store(uint32_t value) { data[index++] = value; }
-  uint32_t Get(size_t n) {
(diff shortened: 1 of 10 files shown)


tcal-x (approved):
I like the namespace.

proj/hps_accel/src/software_cfu.h
```diff
@@ -64,16 +64,17 @@ class Storage {
     len = num_words;
     index = 0;
   }
-  void Store(uint32_t value) { data[index++] = value; }
-  uint32_t Get(size_t n) {
-    if (n == 0 && index >= len) {
+  void Store(uint32_t value) {
+    data[index++] = value;
+    if (index >= len) {
       index = 0;
```
alanvgreen, line 70: Thank you for fixing this.

alanvgreen (approved):
Great!

PR #250, opened 2021-08-31
Update VexRiscv submodule: new "perf" variant;

tcal-x:
"slim" variant updated (larger caches, no perfCSRs)


danc86:
Thanks for doing this tcal, I am going to try out the new slim+cfu variant in hps_accel.

I'm just wondering, what is the "perf" variant for? Its args in the Makefile seem to be identical to the full+cfu variant, or am I misreading it?

tcal-x:
> Thanks for doing this tcal, I am going to try out the new slim+cfu variant in hps_accel.

Yes, that should work, using the current hack of supplying an external variant.   I'm working on a PR to monkey patch the CPU variant info so that we can instead directly ask for slim+cfu.

Be aware that slim+cfu has no perf CSRs, so you need to specify NPROFILE, or the system will hang when it tries to initialize the non-existent CSRs.

> I'm just wondering, what is the "perf" variant for? Its args in the Makefile seem to be identical to the full+cfu variant, or am I misreading it?

You're right, they're the same for now.   Related to what I mentioned above -- I need to get some other stuff in place, and then I can safely remove the perfCSRs from the 'full*' variants, and keep them only in the 'perf' variants.

danc86 (approved):

PR #251, opened 2021-09-02
Enable builds of hps_accel using oxide toolchain

alanvgreen:
Enable builds of hps_accel using oxide toolchain

Reduces clock divisor
Runs many nextpnr-nexus in parallel

Tested with

```
cd proj/hps_accel
time make PLATFORM=hps bitstream
```

For me this runs 54 separate copies of nextpnr and terminates in 14 minutes:

```
make[1]: Leaving directory '/usr/local/google/home/avg/src/CFU-Playground/soc'

real    14m17.820s
user    778m42.951s
sys     1m17.644s
```

### DIFF
--- a/environment
+++ b/environment
@@ -35,3 +35,4 @@ pathadd() {
     fi
 }
 pathadd "${CFU_ROOT}/third_party/usr/local/bin"
+pathadd "${CFU_ROOT}/scripts"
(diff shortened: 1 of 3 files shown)


tcal-x (approved):
Cool!

alanvgreen:
I'm not sure exactly why, but in CI it appears to be failing to find the new file scripts/parallel-nextpnr-nexus.

PR #252, opened 2021-09-02
Separate perf CSRs from NPROFILE flag.

tcal-x:
Now acessing perf CSRs is controlled by a LiteX-defined
flag CONFIG_CPU_PERF_CSRS instead of NPROFILE.
This makes it safe to enable profiling on CPU variants
without perf CSRs (such as fomu and slim).


danc86 (approved):
This is great! We still get the cycle count profile from mcycle, right?

tcal-x:
> This is great! We still get the cycle count profile from mcycle, right?

Exactly, we still get the per-layer and total cycle counts from mcycle.

PR #253, opened 2021-09-04
Monkey patching for CPU variants incl. slim, perf.

tcal-x:
Now there is monkey patching both at the general level, plus more
that is ice40-specific.

In both cases, when updating core.VexRiscv.add_soc_components, 
call the previous version rather than completely replacing it.

### DIFF
--- a/soc/hps_soc.py
+++ b/soc/hps_soc.py
@@ -223,7 +223,7 @@ def main():
     parser.add_argument("--debug", action="store_true",
                         help="Enable debug mode")
     parser.add_argument("--slim_cpu", action="store_true",
-                        help="Use slimmer VexRiscv (required for mnv2_first)")
+                        help="DEPRECATED: use '--cpu-variant=slim+cfu' instead (Use slimmer VexRiscv (required for mnv2_first))")
     parser.add_argument("--build", action="store_true",
                         help="Whether to do a full build, including the bitstream")
     parser.add_argument("--toolchain", default="oxide",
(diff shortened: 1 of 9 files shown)


soc/patch_cpu_variant.py
```diff
@@ -0,0 +1,47 @@
...
+# limitations under the License.
+
+
+# Disable pylint's E1101, which breaks completely on migen
+# pylint:disable=E1101
+
+from litex.soc.cores.cpu.vexriscv import core
+
+def patch_cpu_variant():
+    """Monkey patches custom variants into LiteX."""
```
alanvgreen, line 23: not blocking submit, but in future, please use autopep8 to format Python files.

soc/hps_soc.py
```diff
@@ -248,7 +249,9 @@ def main():
         integrated_rom_init = []
 
     if args.cpu_cfu:
-        if args.slim_cpu:
+        if args.cpu_variant:
+            variant = args.cpu_variant
+        elif args.slim_cpu:
```
alanvgreen, line 254: Why is there both slim_cpu and cpu_variant? Is this supposed to be temporary? If so, perhaps print a message when the old form is used, and also note in the help string for the slim_cpu option.

alanvgreen (approved):

soc/hps_soc.py
tcal-x: Yes, you're right, there's overlapping capability, the old specific version (--slim_cpu) and the new general version (--cpu-variant=slim+cfu).    I should deprecate the old version, and also update any uses of the old version in project Makefiles.

alanvgreen:
Any chance we can get this in soon, with any changes in a follow-up PR? I'd like to use the SlimCFU on proto2. I've already patched it in and it's looking good!

alanvgreen:
Thanks!

PR #257, opened 2021-09-08
Add support for verilated CFU in Renode

robertszczepanski:
This is related to #11.

What's new:
1. New subdirectory `common/renode-verilator-integration` which inludes `CmakeLists.txt`, `sim_main.cpp` and `renode_h.patch`.  These files are used to build verilated CFU library. Since these are adjusted to work as Renode plugin, `renode_h.patch` changes include path in `renode.h` to match our custom structure (we download few files, see 2.).
2. `VerilatorIntegrationLibrary` - files that are part of library are downloaded to `third_party/renode/verilator-integration-library`. These files are minimum to build CFU library and they must be downloaded since these are not present in Renode portable. As soon as Renode includes them in portable version we will get rid of this workaround.
3. Renode scripts are now generating `Verilated.CFUVerilatedPeripheral` with required settings to run it. It's also appended to predefined scripts so new scripts should not contain CFU in them. Our scripts (`proj.mk` with `generate_renode_scripts.py`) will take care if CFU should or should not be included in Renode scripts (if flag `SW_ONLY=1` is passed, CFU won't be added).
4. Projects added to CI are now being tested without any additional build parameters by default. You can add different build variants by adding `ci_build_params.txt.{0-9}` but it won't turn off default tests. It is done for `mnv2_first` right now.
5. `example_cfu` is added to CI workflow.


tcal-x:
Hi @robertszczepanski , I'm trying to run using the `renode-cfu-ci` branch.   I'm encountering an error after Renode starts.   Am I missing some setup?   Here is my command, in `proj/mnv2_first/`:
```
make PLATFORM=common_soc clean renode
```
This is the error:
renode_err

robertszczepanski:
Hi @tcal-x, I can see on attached screenshot that you are using old Renode version. Currently `scripts/setup` won't update it for you but it will print a warning that your local version don't match an expected one. Unfortunately you have to remove `third_party/renode` by hand and then `setup` script will download latest version for you. This should fix your problem.

tcal-x:
Hi @robertszczepanski , I'm now getting this error, which is a bit confusing, since before I updated Renode I was able to complete compilation:
```
make[4]: Entering directory '/home/tim/tcal-x/CFU-Playground/proj/example_cfu/build/renode'
Scanning dependencies of target libVtop
make[4]: Leaving directory '/home/tim/tcal-x/CFU-Playground/proj/example_cfu/build/renode'
make[4]: Entering directory '/home/tim/tcal-x/CFU-Playground/proj/example_cfu/build/renode'
[ 14%] Building CXX object CMakeFiles/libVtop.dir/sim_main.cpp.o
/home/tim/tcal-x/CFU-Playground/common/renode-verilator-integration/sim_main.cpp:17:10: fatal error: src/renode_cfu.h: No such file or directory
   17 | #include "src/renode_cfu.h"
      |          ^~~~~~~~~~~~~~~~~~
compilation terminated.
make[4]: *** [CMakeFiles/libVtop.dir/build.make:104: CMakeFiles/libVtop.dir/sim_main.cpp.o] Error 1
make[4]: Leaving directory '/home/tim/tcal-x/CFU-Playground/proj/example_cfu/build/renode'
make[3]: *** [CMakeFiles/Makefile2:95: CMakeFiles/libVtop.dir/all] Error 2
make[3]: Leaving directory '/home/tim/tcal-x/CFU-Playground/proj/example_cfu/build/renode'
make[2]: *** [CMakeFiles/Makefile2:102: CMakeFiles/libVtop.dir/rule] Error 2
make[2]: Leaving directory '/home/tim/tcal-x/CFU-Playground/proj/example_cfu/build/renode'
make[1]: *** [Makefile:137: libVtop] Error 2
make[1]: Leaving directory '/home/tim/tcal-x/CFU-Playground/proj/example_cfu/build/renode'
make: *** [../proj.mk:177: renode-scripts] Error 2
```

tcal-x:
@robertszczepanski -- nevermind -- my Renode update was incorrect.   I had left an empty `third_party/renode/` directory, which blocked the installation of the new Renode portable.

tcal-x:
@robertszczepanski  -- should this work correctly with any project that uses a CFU?

I'm seeing a golden test failure (menu 1-1-g), using `make PLATFORM=common_soc renode` in `proj/hps_accel/`.   The differences are relatively small such as ` FAIL -127 (actual) != -124 (expected)`.   But the results should be bit-exact.

This test passes on the board, and also passes with Renode when not using the CFU (that is, using the software routines that imitate the CFU functions, by setting `DEFINES += CFU_SOFTWARE_DEFINED`).

robertszczepanski:
@tcal-x do you know if that CFU has anything specific about its logic compared to other projects? It's weird that tests fail and I don't see why it happens yet. I'll investigate it.

tcal-x:
> @tcal-x do you know if that CFU has anything specific about its logic compared to other projects? It's weird that tests fail and I don't see why it happens yet. I'll investigate it.

@danc86 @alanvgreen , we're seeing small errors on `hps_accel` with the Verilator cosimulation.   Is there anything unique about it, to give Robert an idea where to start looking?   `mnv2_first` does work as expected.

robertszczepanski:
@tcal-x 
Actually I've already found a fix to this issue but the change should be applied in `VerilatorPlugin` in Renode. An amount of ticks after receiving `rsp_valid` from CFU is too smal: https://github.com/renode/renode/blob/master/src/Plugins/VerilatorPlugin/VerilatorIntegrationLibrary/src/buses/cfu.cpp#L65. The least number of ticks seems to be `4` to pass the tests.

I checked waveforms on Arty and there's always ~5+ ticks after each CFU operation. 
It looks like asserted `rsp_ready` from CFU doesn't mean it really finished an operation and it needs few more ticks to finish. If you know why it happens, please let me know.

tcal-x:
@robertszczepanski , this is a very interesting finding, and something that we should get to the bottom of.

When you say that the CFU needs a "few more ticks to finish", do you mean that the correct result on `rsp_payload_output_0` doesn't appear until a few cycles after `rsp_valid` goes high, or do you mean that the CFU needs a few cycles to finish its internal computation (and that issuing a new CFU instruction corrupts the ongoing computation)?

robertszczepanski:
I think it's about internal computation. Correct result doesn't appear after few cycles but a wrong result appears immediately with `rsp_valid`.

CFU `functionID`, `ready`/`valid`, `input`/`output` signals look exactly the same (in Renode and on Arty). The only difference is that on real hardware there are few ticks after each CFU operation.

First difference in values I spotted is an output value so it looks a lot like wrong computation earlier. Adding these 3 ticks after each CFU operation makes this problem disappear and output provides proper values.

PiotrZierhoffer:
@tcal-x I wonder how we should approach this. We can adapt, but it also seems to indicate something to be wrong on the verilog side.

tcal-x:
> @tcal-x I wonder how we should approach this. We can adapt, but it also seems to indicate something to be wrong on the verilog side.

Yes, let's not gloss over it, let's understand what's going on.

tcal-x:
To be clear, if correct data shows up on `rsp_payload_output_0` a few cycles after `rsp_valid` goes high, it should be ignored.   The only data that matters is what is on `rsp_payload_output_0` at the clock posedge when `rsp_valid` and `rsp_ready` are high -- only that value should be written back to the CPU register file.

@robertszczepanski , can you post the waveforms at the CFU boundary that you were describing?

robertszczepanski:
@tcal-x - Here's a link to a drive with waveforms: https://drive.google.com/drive/folders/1iLuKl-cgI8Gf5qA6waid_kou89KUsFQ1?usp=sharing
There are 3 waveforms, one from Arty and two from Renode (with and without "fix"). All of them captured on tests from `hps_accel` project menu.
You can see that values are always the same on CFU lines when `cmd_ready` and `cmd_valid` are asserted. However first value on `rsp_payload_outputs_0` is different for `renode_1_tick.vcd`.

tcal-x:
[Different topic: @alanvgreen , is one cycle of reset sufficient, and is the CFU ready for use immediately after the reset?  https://github.com/renode/renode/blob/master/src/Plugins/VerilatorPlugin/VerilatorIntegrationLibrary/src/buses/cfu.cpp#L73-L79 ]

tcal-x:
@robertszczepanski , thank you for the waveforms.  I understand now.   Originally, Verilator simulation paused as soon as the asserted `rsp_valid` signal was seen, and then resumed with the following custom instruction causing `cmd_valid` signal to go high.   So from the point of view of the CFU, almost immediately after finishing one interaction, it was hit with the next one.

With  your fix, the interaction is more like what actually occurs --- there are a few cycles between one custom instruction's response `rsp_valid` and the next custom instruction's initiation `cmd_valid`.

However, even with the original interaction timing, the interface code was correct, that is, the CPU side of the interface was obeying the protocol.   If the CFU was not ready to handle the next interaction, it should have dropped its `cmd_ready` signal, but it didn't.

So, I would like us to figure out what is happening in the CFU.   Also, simulation would be a bit faster without the extra cycles, right?   So let's **not** add the extra cycles now.

It looks like other PR merges have caused a conflict -- can you address that, Robert?

alanvgreen:
@robertszczepanski my version of gtkwave (GTKWave Analyzer v3.3.104) is not working with arty.vcd.

However, looking at renode_1_tick.vcd, I see the CFU responses are being recorded on negative clock edges. Is that what you expected? I'm not seeing anything else unusual in renode_1_tick.vcd or renode_4_ticks.vcd. I'll see if I can get some data out of arty.vcd to compare it with.

The hps_accel accelerator doesn't do anything particularly strange compared to mnv2_first.

alanvgreen:
The first few lines of arty.vcd were:

```
$date
	2021-09-16 16:48
$end
$timescale 1ps $end
$scope 1ps $end
```

When modified to:
```
$date 2021-09-16 16:48
$end
$timescale 1ps $end
$scope module TOP $end
```

I could open the file.

alanvgreen:
@tcal-x: regarding reset, we haven't really thought about it. Any advice you have would be gratefully received. Have raised #277. Let's put it on the list for our weekly meeting.

robertszczepanski:
@tcal-x I resolved conflicts.
@alanvgreen I also noticed that CFU responses are being recorded on negative edge but this is the closest to real hardware model I managed to achieve and results look to be correct anyway. I tried executing `evaluateModel();` in some additional places or changing its current order but it always breaks a model and CFU often doesn't respond at all which ends with `CfuTimeout` error. As long as it returns the same results as real hardware I think we don't need to be bothered about that.

tcal-x (approved):
LGTM

thanks @robertszczepanski !

PR #266, opened 2021-09-14
Various CSR related changes

alanvgreen:


proj/hps_accel/Makefile
```diff
@@ -52,6 +52,9 @@ DEFINES += INCLUDE_MODEL_HPS
 RUN_MENU_ITEMS=1 1 c
 TEST_MENU_ITEMS=3 a
 
+# Ensure perf_csrs logic is present
+DEFINES += CONFIG_CPU_PERF_CSRS
```
danc86, line 56: Hmm, I thought Litex was supposed to define this if the perf CSRs are implemented in our design? If that's not happening, we should fix Litex -- or possibly we don't actually have the perf CSRs :-)

common/src/perf.cc
```diff
@@ -44,19 +44,27 @@ void perf_reset_all_counters() {
...
-    printf("---------+--------+--------+---------+--------------\n");
-  }
+
+  printf(" Counter |  Total | Starts | Average |     Raw\n");
+  printf("---------+--------+--------+---------+--------------\n");
   for (int i = 0; i < NUM_PERF_COUNTERS; ++i) {
     unsigned total = perf_get_counter(i);
     unsigned starts = perf_get_start_count(i);
 
+    // Adjust for overmeasurement per start
```
danc86, line 63: Looks like this may conflict with PR#264?

proj/hps_accel/Makefile
tcal-x: The setting of CONFIG_CPU_PERF_CSRS is in a WIP PR I have.   It will be added as a monkey patch applied to LiteX.   So yes, this needs to be done for now, and can / should be removed once that PR lands (#253).

common/src/perf.cc
alanvgreen: All tidied up now.

danc86 (approved):

PR #269, opened 2021-09-15
hps_accel: tighter inner loop

danc86:


proj/hps_accel/src/blocks.cc
```diff
@@ -77,10 +77,7 @@ Vector16 Vector16::build(int8_t a, int8_t b, int8_t c, int8_t d, int8_t e,
 
 Vector16 Vector16::zeroes() { return Vector16{{0, 0, 0, 0}}; }
 
-// Performs a 16 x 16 vector multiplication
-int32_t multiply_accumulate() {
-  return cfu_get(REG_MACC_OUT);
-}
+extern int32_t multiply_accumulate();
```
alanvgreen, line 80: Why is this needed? Are there link errors without it?

alanvgreen (approved):

proj/hps_accel/gateware/get.py
```diff
@@ -139,12 +156,18 @@ def read_reg(self, m, registers, register_num,
...
+                                       ready_when_valid=(i not in self.READ_EXACTLY_ONCE))
                      for i in self.REGISTER_IDS}
         for i, register in registers.items():
             m.submodules[f"reg_{i:02x}"] = register
             m.d.comb += connect(self.input_streams[i], register.input)
-            m.d.comb += register.invalidate.eq(self.invalidates[i])
+            if i in self.READ_EXACTLY_ONCE:
+                # Reading from it invalidates the value, in preparation for
+                # receiving the next one.
+                m.d.comb += register.invalidate.eq(self.read_strobes[i])
```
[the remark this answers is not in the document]
danc86: Yes good point

proj/hps_accel/src/blocks.cc
danc86: It ensures there is a symbol defined for the inline function. I originally saw this rule in the Opentitan style guide:

https://docs.opentitan.org/doc/rm/c_cpp_coding_style/#:~:text=compilation%20unit%20with%20a%20compatible%20extern%20declaration

and I assumed there was a similar rule in the Google C++ style guide, although I see now there isn't.

We don't actually need these. Although these inline functions *are* exported, they are only exported to the other files in our project here, so if we did need a real symbol defined for them we would know about it.

I'll remove them.

danc86:
Amended to:
* also obey the `invalidates` signal for the read-exactly-once registers
* remove the dead test
* remove the unneeded `extern` definitions

proj/hps_accel/src/blocks.cc
alanvgreen: Oh, I see it matches C11. That's a good thing to keep in mind.

alanvgreen:
LGTM.

PR #272, opened 2021-09-16
hps_accel/mem: new WideReadMemory class

alanvgreen:
A simpler memory that is written one 32 bit word at a time and read 4
words (128bits) at a time.

### DIFF
--- a/proj/hps_accel/gateware/mem.py
+++ b/proj/hps_accel/gateware/mem.py
@@ -1,0 +1,75 @@
+# Copyright 2021 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     https://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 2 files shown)


proj/hps_accel/gateware/mem.py
```diff
@@ -0,0 +1,75 @@
...
+    read_addr: Signal(range(depth / 4)), in
+      The address to read four words from. Lower two bits of words address
+      are not present since all four words are retrieved.
+    read_data: Signal(128), out
+      The four 32 bit words read from memory. Data is available the cycle
+      after the read_addr is placed on the bus
+    """
+
+    def __init__(self, depth):
+        self.__depth = depth
```
danc86, line 50: Why the double-underscored attribute name?

I guess you probably know that single-underscore is conventional for private attributes, double-underscore has the surprising behaviour with name mangling.

Doesn't really matter either way in this case though since we don't expect this to be subclassed.

danc86 (approved):

danc86:
Also, thanks for whipping this up so fast. :-) I'm going to give it a go in my `InputStore` patch now.

danc86:
One of the Renode tests failed with a timeout after 5m 13s, I guess it's just a flake? I've clicked retry to see if it fixes itself.

proj/hps_accel/gateware/mem.py
alanvgreen: Oops. That is showing my old-school Python roots.

PR #273, opened 2021-09-16
Renode expect support

### LINKED ISSUE #271: Renode: there should be a single-command to test in Renode
When I test on arty, I can run

```
$ make unit
```

Which will load the C code, choose some menu items then exit back to the command line.

It would be nice if there were an equivalent with renode, as it would allow faster iteration. 

I seem to recall that there used to be something like this for renode? Or am I misremembering?

PiotrZierhoffer:
This adds 3 targets:

- renode-headless, which just runs Renode in a single console
- unit-renode, same as `make unit` but running in Renode
- run-renode, similar to `make run`, added for completeness

All of these accept the `TARGET` variable to select the actual target.

Closes #271

### DIFF
--- a/proj/proj.mk
+++ b/proj/proj.mk
@@ -164,10 +164,14 @@ endif
 
 TARGET_REPL := $(BUILD_DIR)/renode/$(TARGET)_generated.repl
 
-.PHONY:	renode 
+.PHONY:	renode
 renode: renode-scripts
 	@echo Running interactively under renode
 	pushd $(PROJ_DIR)/build/renode/ && $(RENODE_DIR)/renode -e "s @$(TARGET).resc" && popd
+
+.PHONY:	renode-headless
+renode-headless: renode-scripts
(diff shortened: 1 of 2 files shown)


proj/proj.mk
```diff
@@ -273,6 +277,14 @@ unit: $(SOFTWARE_BIN)
 	@echo Running unit test on board
 	$(BUILD_DIR)/interact.expect $(SOFTWARE_BIN) $(TTY) $(UART_SPEED) $(TEST_MENU_ITEMS) |& tee $(UNITTEST_LOG)
 
+run-renode: $(SOFTWARE_ELF) renode-scripts
+	@echo Running automated test in Renode
+	$(BUILD_DIR)/interact.expect r $(RUN_MENU_ITEMS) |& tee $(UNITTEST_LOG)
```
tcal-x, line 282: Should be $(SOFTWARE_LOG)

PiotrZierhoffer: Good catch, fixed

tcal-x (approved):
LGTM

alanvgreen:
Thanks!

PR #274, opened 2021-09-16
Unify and reduce wget output in the whole project

PiotrZierhoffer:
Right now wget is extremely noisy or completely silent (in Dockerfile).

It makes sense to have some output indicating the progress, but frankly
the default settings make CI logs unreadable.

This reduces the output so that each line covers 32M of downloaded data.


tcal-x (approved):
Thanks Piotr.

PR #278, opened 2021-09-18
hps_accel: accelerate MultiplyByQuantizedMultiplier

alanvgreen:
Accelerates the MultiplyByQuantizedMultiplier operation over the range of parameters we see in the current hps model.


proj/hps_accel/src/blocks.cc
```diff
@@ -161,4 +158,73 @@ int32_t MultiplyByQuantizedMultiplier_03(int32_t x,
...
+      mask = (1 << 10) - 1;
+      break;
+    case -11:
+      quotient = product >> 11;
+      mask = (1 << 11) - 1;
+      break;
+    default:
+      quotient = 0;
+      mask = 0;
+      printf("***BAD SHIFT\n");
```
danc86, line 222: Consider also DCHECK'ing the sanity of this shift value (and other expectations) instead of, or in addition to, this printf.

Currently a DCHECK failure just hangs the whole CPU but I was intending at some point to make our assert() and abort() implementation friendlier and print some good messages before it dies.

I'm regularly running through the model without NDEBUG (meaning, do the assertions) to catch mistakes.

proj/hps_accel/gateware/math.py
```diff
@@ -0,0 +1,131 @@
...
+
+        # Cycle 0: register inputs
+        a = self.input.payload.a
+        reg_a = Signal(signed(32))
+        reg_b = Signal(signed(32))
+        m.d.sync += reg_a.eq(Mux(a >= 0, a, -a))
+        m.d.sync += reg_b.eq(self.input.payload.b)
+
+        # Cycle 1: multiply to register
+        # both operands are positive, so result always positive
```
danc86, line 88: Should the `reg_ab` (and `reg_a` and `reg_b`) signals not be unsigned then? Or do we still need a sign bit for some reason?

proj/hps_accel/gateware/math.py
```diff
@@ -0,0 +1,131 @@
...
+        reg_ab = Signal(signed(63))
+        m.d.sync += reg_ab.eq(reg_a * reg_b)
+
+        # Cycle 2: nudge, take high bits and sign
+        positive_2 = self.delay(m, 2, a >= 0)
```
danc86, line 93: Tricky

proj/hps_accel/gateware/math.py
```diff
@@ -0,0 +1,131 @@
...
+        m.d.sync += self.output.payload.eq(with_sign)
+
+
+class MathInstruction(InstructionBase):
+    """An instruction used to perform specialist math operations.
+
+    This instruction is likely to be replaced with more comprehensive
+    gateware in future.
+
+    Sets a configuration register from in0.
```
danc86, line 167: I don't grok this sentence in this context, is it actually a copy-paste error?

danc86 (approved):
Very cool

proj/hps_accel/gateware/math.py
alanvgreen: Thanks. Yes. That is what happened. Fixed.

proj/hps_accel/gateware/math.py
alanvgreen: Sorry :(. There's not really any great choices for expressing "take a value from two cycles ago". In the next PR, there is a slightly better comment here.

proj/hps_accel/gateware/math.py
alanvgreen: I think you're right, but the least significant bit of the result is dependent on a some math that I don't fully understand. I'd be 100% OK with changes to simplify this code further.

proj/hps_accel/src/blocks.cc
alanvgreen: Done. Thank you for the suggestion. I guess we should also do DCHECK checking in the CI.

alanvgreen (commented):
Thanks for the quick review!

PR #279, opened 2021-09-20
fix fat finger issue

alanvgreen:


danc86 (approved):

alanvgreen:
Well this is thoroughly broken - the github CI uses an older version of bash that does not understand "wait -p".

I'm supposing the shell script should be rewritten in Python.

PR #280, opened 2021-09-20
Build for proto2

alanvgreen:
A set of patches to allow building hps_accel for PLATFORM=hps

Significant changes:

1. use parallel-nextpnr-nexus to get at least one good build
2. use slim/perf/cfu variant
3. use untiled model and reduce memory usage to suit
4. Dramatically reduce software_cfu memory use. This breaks software_cfu on hps_accel, but this is currently not a problem.


tcal-x:
@alanvgreen ?   Is this all obsolete now?

alanvgreen:
Yes. Let's close it.

PR #283, opened 2021-09-20
common/Makefile: fix software.elf.sym

alanvgreen:
software.elf.sym was being generated with -s, which caused the entire
content of the elf file to be translated into hex. This fix uses -t
instead which just includes symbol positions and lengths.


danc86 (approved):
I've made this mistake more than once :-) it's `readelf -s` but `objdump -t`, good luck getting my brain to ever remember that

tcal-x (approved):
LGTM

PR #287, opened 2021-09-21
CI: run no CFU, trivial CFU, real CFU builds with Oxide (both Conda and fresh-built).

tcal-x:
15 minute timeout on each run, error ignored.


.github/workflows/fmax-trials.yml
```diff
@@ -19,12 +19,12 @@ jobs:
       - run: which pip3 && which python3 && which pip
       - run: riscv64-unknown-elf-gcc --version
       - run: bash scripts/install_oxide
```
alanvgreen, line 21: I wonder whether we could use a prebuilt oxide here? Maybe a nightly?

tcal-x: I already have a "make-env" branch using Conda packages; it's waiting for a suitable RISC-V toolchain package.   But I suppose we can mix-and-match, using a SiFive toolchain tarball and everything else from Conda packages.

mithro:
Key poking Antmicro about the compiler toolchain. It should have already been finished by now. Ping @PiotrZierhoffer ?

.github/workflows/fmax-trials.yml
alanvgreen: That would definitely speed up the CI.

tcal-x:
Now a 5th arg is passed to the parallel script, the user-specified random seed.   This is only used when the script falls back to the non-parallel nextpnr run.   It allows the CI workflow to specify different random seeds for the multiple runs.

scripts/parallel-nextpnr-nexus
```diff
@@ -37,7 +38,7 @@ if ! [[ ${BASH_VERSION} =~ "5.1" ]]; then
         --pdc "${PDC}" \
         --fasm "${RESULT_FASM}" \
         --device LIFCL-17-8UWG72C \
-        --seed 1
+        --seed "${SEED}"
```
alanvgreen, line 41: I'd be happy to use `seed_base = $((SEED * 100))` instead of setting seed_base randomly below.

(This shouldn't block merge - it's something I could do later)

alanvgreen (approved):

PR #289, opened 2021-09-21
hps_accel: fix the on-device tests for the macc unit

danc86:
This line was accidentally dropped in commit 7fd3e19a. We need it due to
commit 8f308c35.


alanvgreen (approved):
Oops. Thank you.

Let's discuss whether we can put this kind of thing into the CI.

PR #292, opened 2021-09-22
hps_accel: add post processing pipeline

alanvgreen:
This builds on the previous SRDHM and RDBPOT instructions, adding a whole pipeline for the functionality.


danc86 (approved):
I also went back and studied the previous PR, and this one, to see if I could spot where the timing sensitivity could be in the previous PR (and how this could have fixed it). I couldn't see anything though. So that makes me nervous...

PR #291, opened 2021-09-21
Merge upstream VexRiscv and Litex.

tcal-x:
VexRiscv now based on SpinalHDL 1.6.0.
Signal 'rsp_payload_response_ok' has been removed.

Signed-off-by: Tim Callahan <tcal@google.com>

PR #294, opened 2021-09-22
bump litedram and reduce hps_accel on Arty to 75MHz

danc86:
@tcal-x is there anything else I should test for the litedram bump, apart from that my designs are still working fine on Arty?


tcal-x:
Hi @danc86 ; I have the litedram bump also in #291, but not the clock rate adjustment.

I don't think the oxide test here will pass; it's stuck on the `wait -p` issue.

tcal-x (approved):
LGTM

PR #297, opened 2021-09-23
hps_accel: single Conv2D test

alanvgreen:
Tests a single Conv2D op, based on actual data from the HPS model. A
single layer is easier to work with in some kinds of debugging.

Includes code to dump a layer's input and output data to the console, as
well as infrastructure to use that dumped data in a test.

This test will not work on the proto2 board due to memory size
constraints. In future, we may wish to exclude it from the proto2
binary, but there is currently no simple mechanism to do so.


proj/hps_accel/Makefile
```diff
@@ -19,11 +19,15 @@ export DEFINES :=
...
+# For development purposes, dump a layer in accelerated code
+# (Large effect on performance, much output)
+#DEFINES += DUMP_LAYER=20
+
 # Uncomment this line to use software defined CFU functions in software_cfu.cc
 #DEFINES += CFU_SOFTWARE_DEFINED
 
 # Uncomment this line to skip debug code (large effect on performance)
-DEFINES += NDEBUG
+#DEFINES += NDEBUG
```
danc86, line 30: May not want to land this line

danc86 (approved):

alanvgreen:
How does one "finish a review" in the git hub interface. Is it like this?

proj/hps_accel/Makefile
alanvgreen: Thanks!

PR #300, opened 2021-09-24
hps_accel: new post process pipeline

alanvgreen:
Sends output from the post process pipeline to a FIFO so it can be processed asynchronously.

### DIFF
--- a/proj/hps_accel/Makefile
+++ b/proj/hps_accel/Makefile
@@ -24,10 +24,13 @@ DEFINES += ACCEL_CONV
 #DEFINES += DUMP_LAYER=20
 
 # Uncomment this line to use software defined CFU functions in software_cfu.cc
-#DEFINES += CFU_SOFTWARE_DEFINED
+DEFINES += CFU_SOFTWARE_DEFINED
 
 # Uncomment this line to skip debug code (large effect on performance)
 DEFINES += NDEBUG
+
+# Show output from TF_LITE_ASSERTs
+DEFINES += TF_LITE_MCU_DEBUG_LOG
 
 # Uncomment this line to skip individual profiling output (has minor effect on performance).
 #DEFINES += NPROFILE
--- a/proj/hps_accel/Makefile
+++ b/proj/hps_accel/Makefile
@@ -24,7 +24,7 @@ DEFINES += ACCEL_CONV
 #DEFINES += DUMP_LAYER=20
 
 # Uncomment this line to use software defined CFU functions in software_cfu.cc
-DEFINES += CFU_SOFTWARE_DEFINED
+#DEFINES += CFU_SOFTWARE_DEFINED
 
 # Uncomment this line to skip debug code (large effect on performance)
 DEFINES += NDEBUG
--- a/proj/hps_accel/gateware/post_process.py
+++ b/proj/hps_accel/gateware/post_process.py
@@ -17,6 +17,8 @@ from nmigen_cfu import SimpleElaboratable, InstructionBase
 from nmigen import signed, unsigned, Memory, Module, Mux, Record, Signal
 
 from .stream import BinaryPipelineActor
+from .stream.gearbox import ByteToWord
+from .stream.fifo import StreamFifo
 from .constants import Constants
 from .stream import connect, Endpoint, BinaryPipelineActor
 
@@ -354,6 +356,9 @@ class PostProcessInstruction(InstructionBase):
       Data read from OutputStorageParams
     read_enable: Signal(), out
       Tells OutputStorageParams to read next
+
+    result: Endpoint(unsigned(32)), out
+      Word result from post process pipline
     """
 
     def __init__(self):
@@ -363,9 +368,10 @@ class PostProcessInstruction(InstructionBase):
         self.activation_max = Signal(signed(8))
         self.read_data = Record(OUTPUT_PARAMS)
         self.read_enable = Signal()
+        self.result = Endpoint(unsigned(32))
 
     def elab(self, m: Module):
-        # Connect the post process pipeline
+        # Oost process pipeline and its parameters
         m.submodules.ppp = ppp = PostProcessPipeline()
         m.d.comb += ppp.output.ready.eq(1)
         m.d.comb += [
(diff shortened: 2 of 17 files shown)


proj/hps_accel/src/tensorflow/lite/kernels/internal/reference/integer_ops/conv_accel.cc
```diff
@@ -111,14 +113,25 @@ void ConvPerChannel4x4(const ConvParams& params,
                          MAX_INPUT_WORDS);
         hps_accel::LoadInput(input_width, input_depth, current_input_data);
 
+        // Calculate all outputs for a single output pixel
         for (int out_channel = out_channel_offset;
              out_channel < out_channel_offset + output_channels;
              ++out_channel) {
-          int iterations = filter_height * filter_width * input_depth / 16;
+                         int iterations = filter_height * filter_width * input_depth / 16;
```
danc86: Slightly weird indentation here but you probably fix this in a subsequent commit so no big deal

proj/hps_accel/src/tensorflow/lite/kernels/internal/reference/integer_ops/conv_accel.cc
```diff
@@ -117,22 +120,20 @@ void ConvPerChannel4x4(const ConvParams& params,
...
-             out_channel < out_channel_offset + output_channels;
-             out_channel += 4) {
-          // TODO: less pointer arithmetic
-          uint32_t* output_data_ptr = static_cast<uint32_t*>(
-              static_cast<void*>(output_data + Offset(output_shape, 0, out_y,
-                                                      out_x, out_channel)));
-          *output_data_ptr = hps_accel::GetOutputWord();
+        // Pull result data from output FIFO and place into memory, a word at a
+        // time.
+        uint32_t* output_data32 = output_data32_base;
```
danc86, line 131: I can see how this is a useful optimisation (does it produce a substantial improvement in speed by itself, or is it really just preparation for doing some fancier DMA later?).

But I liked the safety of Offset() doing the pointer arithematic for me, to reduce the likelihood I would get some dimension wrong somewhere. So I'm a bit sad we lose that.

Would it be worth adding a DCHECK inside the loop which cross-checks the pointer against the expected offset? Or I am being too paranoid perhaps?

proj/hps_accel/gateware/stream/gearbox.py
```diff
@@ -0,0 +1,74 @@
...
+
+    def __init__(self):
+        self.input = Endpoint(signed(8))
+        self.output = Endpoint(unsigned(32))
+
+    def elab(self, m: Module):
+
+        # Registers accumulate incoming values
+        registers = [Signal(8, name="register{i}") for i in range(3)]
+        waiting_to_send = Signal()
```
danc86, line 44: Neat. I probably needed to use some pattern like this in the store rewrite that I tried. I have made a note to think about this when I look at the stores again.

proj/hps_accel/gateware/stream/fifo.py
```diff
@@ -0,0 +1,63 @@
...
+
+from nmigen import signed, unsigned, Cat, Signal, Module
+from nmigen.lib.fifo import SyncFIFOBuffered
+from nmigen.hdl.dsl import FSM
+from nmigen_cfu import SimpleElaboratable
+
+from .stream import PayloadDefinition, Endpoint
+
+
+class StreamFifo(SimpleElaboratable):
```
danc86, line 25: Hmm we can add this to the list of stuff that should be upstreamed in nmigen streams I guess :-)

proj/hps_accel/gateware/post_process.py
```diff
@@ -363,9 +368,10 @@ def __init__(self):
         self.activation_max = Signal(signed(8))
         self.read_data = Record(OUTPUT_PARAMS)
         self.read_enable = Signal()
+        self.result = Endpoint(unsigned(32))
 
     def elab(self, m: Module):
-        # Connect the post process pipeline
+        # Oost process pipeline and its parameters
```
danc86, line 374: s/Oost/Post/ :-)

proj/hps_accel/Makefile
```diff
@@ -24,7 +24,7 @@ DEFINES += ACCEL_CONV
 #DEFINES += DUMP_LAYER=20
 
 # Uncomment this line to use software defined CFU functions in software_cfu.cc
-DEFINES += CFU_SOFTWARE_DEFINED
+#DEFINES += CFU_SOFTWARE_DEFINED
```
danc86: This commit's contents don't seem to match its message, I guess a minor rebasing mistake?

danc86 (approved):
Woohoo ⚙️⚙️⚙️

proj/hps_accel/src/tensorflow/lite/kernels/internal/reference/integer_ops/conv_accel.cc
alanvgreen: Yes. I double checked that it is fixed.

proj/hps_accel/src/tensorflow/lite/kernels/internal/reference/integer_ops/conv_accel.cc
alanvgreen: Offset() is rather slow, so moving it outside the loops seems best.

The "safety" of Offset() is rather dubious when DCHECKs are off.

I'll add a DCHECK in a follow up PR.

proj/hps_accel/gateware/stream/fifo.py
alanvgreen: Yup! The whole stream directory is candidate for being reusable somewhere.

proj/hps_accel/Makefile
alanvgreen: I'll reword the commit message

proj/hps_accel/src/tensorflow/lite/kernels/internal/reference/integer_ops/conv_accel.cc
danc86: Yes it's only safety in the sense that when I come along later and try to change this code, I would be able to turn off NDEBUG and satisfy myself I haven't messed up the pointer arithmetic.

proj/hps_accel/src/tensorflow/lite/kernels/internal/reference/integer_ops/conv_accel.cc
alanvgreen: Actually, it started to bother me, so I fixed it.

PR #307, opened 2021-09-30
hps: router1, no abc9, 64MHz

danc86:
This is the configuration I am using to get numbers in our tracking spreadsheet.

Alan, you may prefer to keep the parallel router2 runs by default? I found it unnecessary but maybe it gives better results? I note that the parallel-nextpnr script doesn't look for a seed that gives the *best* result, just whichever one completes first.


tcal-x (approved):
LGTM

danc86:
I'll merge this now so that we have some consistency between CFU-Playground and our hps-firmware.git tree, but we can always change this again later.

PR #308, opened 2021-09-30
hps: switch SPI flash to DDR mode

danc86:


danc86:
FYI @kgugala this works nicely on HPS, and gives a reduction in resource usage. But I confirmed this only reduces model evaluation time by ~0.35%.

kgugala (approved):
LGTM

kgugala:
> FYI @kgugala this works nicely on HPS, and gives a reduction in resource usage. But I confirmed this only reduces model evaluation time by ~0.35%.

is this with picolibc and LTO?

danc86:
Yes, these are the numbers I have in my spreadsheet running on HPS proto2 board, evaluating HPS model.

commit 10400f2f7fe0eafc32ed3c1f42d2f46aa553fb96 (last week)
= 193473536 cycles

commit 740150c96dc8e511f1254750bb4adf212ae3f414 (current main)
= 176910336 cycles

this PR
= 176288768 cycles

It suggests picolibc and LTO give a good speed improvement, as you reported, but effectively no change from doubling the SPI flash clock. It doesn't seem right.

kgugala:
maybe to biggest bottleneck is not in flash now?

enjoy-digital:
@danc86, @kgugala: I'd like to merge https://github.com/litex-hub/litespi/pull/60 that simplifies DDR mode but need feedback (I was only able to test on Arty), could you do a test with it? Thanks.

tcal-x (approved):

danc86:
> @danc86, @kgugala: I'd like to merge [litex-hub/litespi#60](https://github.com/litex-hub/litespi/pull/60) that simplifies DDR mode but need feedback (I was only able to test on Arty), could you do a test with it? Thanks.

I tried with that PR applied and it *doesn't* work (CPU never boots from SPI flash). I haven't looked closer at why not. I'll write a comment with more details on that Litespi PR.

danc86:
I want to do more investigation on this (including to understand why there is no effective difference in model evaluation time) but I think we should merge it now and continue from here.

PR #310, opened 2021-10-01
Remove pin from nextpnr-nexus since default 'main' is now correct.

tcal-x:


PiotrZierhoffer (approved):
lgtm!

PR #315, opened 2021-10-06
hps_accel: use slim CPU

danc86:
This buys us a lot more breathing room in terms of avoiding FPGA
congestion, at a cost of about 35% slower model evaluation time.


tcal-x (approved):
LGTM

PR #316, opened 2021-10-06
common/Makefile: sort find output

piotr-binkowski:
This PR modifies the Makefile used for building software to sort found sources.
This ensures that the object files will be linked in the same order on all machines - without that binaries built on different hosts can get different model evaluation times.


tcal-x (approved):
LGTM

PR #318, opened 2021-10-06
cfu: generate rsp_valid without using cmd_valid

piotr-binkowski:
This PR removes combinational path from `cmd_valid` to `rsp_valid` which slightly improves fMax (1-2MHz in recent tests) and FWIW doesn't impact model evaluation time.


tcal-x:
@piotr-binkowski  what are the implications of this change?   Combinational CFU operations are no longer allowed?   Or were they not allowed previously?

PR #323, opened 2021-10-12
Optimize function layout in hps_accel project

j-piecuch:
This PR makes the `LDSCRIPT` make variable overridable from a project, and uses a project-specific linker script in `hps_accel` that prevents instruction cache conflicts between `ConvPerChannel4x4()` and `LoadInput()`. It also fixes a small bug in `hps_accel/Makefile`.

### DIFF
--- a/proj/hps_accel/linker.ld
+++ b/proj/hps_accel/linker.ld
@@ -12,6 +12,8 @@ SECTIONS
 	{
 		_ftext = .;
 		*(.text.start)
+    /* Place the ConvPerChannel4x4() and LoadInput() functions sequentially
+     * to prevent one from evicting the other from the instruction cache. */
     */conv_accel.o(.text.*ConvPerChannel4x4*)
     */blocks.o(.text.*LoadInput*)
 		*(.text .stub .text.* .gnu.linkonce.t.*)


j-piecuch:
@danc86 @tcal-x apologies, I was not aware of the existing overlay system.
In that case, we should probably come up with some other way of letting projects override the order of functions in the `.text` section.

One simple, but somewhat hacky solution is to use `sed` to patch the linker script once it has been copied to the build directory. For this to work, each linker script would need to include some specific string that tells us where to insert the overrides.
For instance, the `.text` section in `common/ld/linker.ld` script could look like this:
```
	.text :
	{
		_ftext = .;
		*(.text.start)
		/* TEXT_OVERRIDES */
		*(.text .stub .text.* .gnu.linkonce.t.*)
		_etext = .;
	} > main_ram
```

In the project's Makefile, we could define a make variable that contains the overrides:
```
TEXT_OVERRIDES := */conv_accel.o(.text.*ConvPerChannel4x4*) */blocks.o(.text.*LoadInput*)
```
We can perform the substitution in the recipe for `build-dir` in `proj.mk`: 
```
.PHONY: build-dir
build-dir: $(BUILD_DIR)/src tflite-micro-src $(BUILD_DIR_EXTRA_DEP) 
	@echo "build-dir: copying source to build dir"
	$(COPY) $(COMMON_DIR)/*              $(BUILD_DIR)
	$(COPY) $(MLCOMMONS_SRC_DIR)/*       $(BUILD_DIR)/src
	$(COPY) $(SAXON_SRC_DIR)/riscv.h     $(BUILD_DIR)/src
	$(COPY) $(SRC_DIR)/*                 $(BUILD_DIR)/src
	$(RM)			             $(BUILD_DIR)/_*
# Overlay platform / target specific changes.
ifneq ($(wildcard $(COMMON_DIR)/_$(PLATFORM)/$(TARGET)/*),)
	$(COPY) $(COMMON_DIR)/_$(PLATFORM)/$(TARGET)/* $(BUILD_DIR)
endif
# Allow project makefile to override placement of functions in the .text section
	sed -i -e '/.*TEXT_OVERRIDES.*/c\$(TEXT_OVERRIDES)' $(BUILD_DIR)/ld/linker.ld

```

A somewhat less hacky way to accomplish the same goal would be to use some macro/templating engine like jinja or the C preprocessor.

PR #329, opened 2021-10-20
Use Conda to get Oxide tools in CI

tcal-x:
(instead of building from scratch each time).

### DIFF
--- a/.github/workflows/oxide.yml
+++ b/.github/workflows/oxide.yml
@@ -12,7 +12,9 @@ jobs:
           python-version: '3.7'
       - run: bash scripts/setup -ci
       - run: which pip3 && which python3 && which pip
-      - run: riscv64-unknown-elf-gcc --version
       - run: make env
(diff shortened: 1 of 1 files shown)


tcal-x:
> Do we have any easy way to control (or even just to see) which git revision of yosys/nextpnr/prjoxide we get from Conda? Or is it entirely at the whim of the Conda package maintainers, or always tracks upstream master branch, or...?

@danc86 yes we can control it, but I don't currently (we just take the latest package tagged with "main").   They get tagged with "main" when they pass their independent regression tests, but they're not tested together, so things could possibly break here.

Here's an example where they control which version they get:
https://github.com/SymbiFlow/fpga-tool-perf/blob/master/conf/nextpnr/environment.yml 

I will add a printout of the version hashes of yosys and nextpnr-nexus to the workflow here.   I already do that in the "fmax" workflow: https://github.com/google/CFU-Playground/blob/main/.github/workflows/fmax-trials.yml#L20

PR #332, opened 2021-10-26
Add ice40 and riscv toolchain support to the conda env;

tcal-x:
add Fomu GHA that uses new conda-provided gcc.


tcal-x:
@alanvgreen PTAL.   This doesn't make the final jump to Conda but it does use Conda to acquire Ice40 tools for running in CI.

alanvgreen (approved):
LGTM.

PR #334, opened 2021-10-29
proj/hps_accel: move gateware to gen1 subdirectory

alanvgreen:
Moves gateware to a sub-directory gateware/gen1 in anticipation of there
being a gen2 sub-directory in future.


danc86 (approved):
This makes me a little sad -- I was still hoping to find time to go back and finish off / experiment further with all the remaining ideas for this original design, to see how fast I could get it.

I guess I can still do that, if we're keeping gen1 in here like this :-)

alanvgreen:
Yeah, I'm not 100% sold on gen2 yet - a bunch of unknowns remain.

PR #335, opened 2021-11-09
Add new variant "slimopt+cfu": simpler handshake, larger i-cache.

tcal-x:
slim+cfu is still the default variant used with hps_accel.   To get slimopt+cfu, add this to the `make` command line: `EXTRA_LITEX_ARGS="--cpu-variant=slimopt+cfu"`, or edit the line in `proj/hps_accel/Makefile`.

I see consistently better clock frequency with the `slimopt+cfu` variant for `hps_accel` using the default `router1` -- MHz in the 80s rather than in the 70s.


alanvgreen:
What is the effect of setting CFU response always ready? Will that allow the toolchain to simplify the gateware?

tcal-x:
> What is the effect of setting CFU response always ready? Will that allow the toolchain to simplify the gateware?

It is the CPU indicating that it is always ready to accept the CFU's response.   The CFU is guaranteed that when it asserts response valid, the payload will be accepted immediately.

In the words of Charles: 

> CFU_FLOW_RESP_READY_ALWAYS mean that the CPU will never put the ready of the response stream to 0, in other words, the CPU will always accept CFU responses.
>
> It relax the CFU side of things, but cost some buffers in the CPU.

alanvgreen:
Hmmm. That would allow us to remove some buffers on the gateware side. I'll keep this in mind. Thanks.

alanvgreen (approved):

PR #337, opened 2021-11-14
Restore the use of -abc9 with the HPS platform.

tcal-x:


tcal-x:
This PR (turning on -abc9) with the current `hps_accel` reduces `OXIDE_COMB` count by 1.9%.

danc86 (approved):
Thanks @tcal-x .

I tried the latest yosys master branch in hps-firmware this morning, and although ABC9 appears to work well there, router1 never converges (at least with the default seed 1). 5278 seconds and still 10118 unsolved arcs. :-(

I guess ABC9 may give a slight improvement to resource usage but it either makes routing slightly harder, or just perturbs things such that seed 1 no longer works.

PR #338, opened 2021-11-14
Gen2a

alanvgreen:
Next set of changes for gen2 accelerator. Moving toward being able to build a test that integrates a systolic array and post processor.

### DIFF
--- a/proj/hps_accel/gateware/gen2/sysarray.py
+++ b/proj/hps_accel/gateware/gen2/sysarray.py
@@ -1,0 +1,133 @@
+# Copyright 2021 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     https://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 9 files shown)


proj/hps_accel/gateware/gen2/sysarray.py
```diff
@@ -0,0 +1,133 @@
...
+        accumulator may be sufficient, and will make more efficient
+        use of the FPGA fabric.
+
+    Attributes
+    ----------
+
+    a_inputs: [Signal(n * a_shape.width)], in
+      List of words forming the A inputs
+
+    b_inputs: [Signal(32)], in
```
danc86, line 61: I guess this is actually a list of `Signal(n * b_shape.width)` which in our case is `Signal(32)`, right?

danc86 (approved):
Exciting

proj/hps_accel/gateware/gen2/sysarray.py
tcal-x: @alanvgreen why the difference between how `a_inputs` and `b_inputs` are declared?

alanvgreen: No good reason. Fixed.

tcal-x: @alanvgreen did you push your change, or am I misunderstanding something?

alanvgreen: fixed now

PR #340, opened 2021-11-15
docs: upgrade to Sphinx 4.3

danc86:
Trying to fix breakage in CI due to docutils 0.18 incompatibility.

### DIFF
--- a/docs/requirements.txt
+++ b/docs/requirements.txt
@@ -1,5 +1,5 @@
-Sphinx==3.3.0
+Sphinx==4.3.0
 sphinx-material==0.0.32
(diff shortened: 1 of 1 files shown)


danc86:
The newer Sphinx works fine, and more importantly it constrains the version of docutils so that pip picks 0.17 as desired:

```
Collecting docutils<0.18,>=0.14
  Downloading docutils-0.17.1-py2.py3-none-any.whl (575 kB)
```

which lets the check-docs job pass in CI.

tcal-x (approved):
LGTM

PR #341, opened 2021-11-16
Add --separate-arena option to create tensor arena.

tcal-x:
4 LRAMs are used for the arena, 1 LRAM for main memory.
The arena is created at 0x60000000.

Testing: 
* It built with the `hps_accel` CFU gateware and met timing (80MHz+).
* Exercised on the proto2 board using BIOS.

Software has *not* been altered to use the arena.

### DIFF
--- a/soc/hps_soc.py
+++ b/soc/hps_soc.py
@@ -61,10 +61,12 @@ class HpsSoC(LiteXSoC):
     # that.
     rom_offset = 2*MB
     sram_origin = 0x40000000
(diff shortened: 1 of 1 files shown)


danc86 (approved):

tcal-x:
I also tested it on the proto2 board with the CFU-Playground software, built with just the KWS model so that it would fit into the 64kB main memory.   Golden tests and unit tests passed.

PR #342, opened 2021-11-18
Update software/linker to use separate tensor arena region.

tcal-x:
To use, add this to your make command:
EXTRA_LITEX_ARGS="--separate-arena"

Tested on proto2 both with and without this new option.

### DIFF
--- a/soc/hps_soc.py
+++ b/soc/hps_soc.py
@@ -107,8 +107,7 @@ class HpsSoC(LiteXSoC):
             ram_size = RAM_SIZE
             arena_size = 0
         self.setup_ram(size=ram_size)
-        if arena_size > 0:
-            self.setup_arena(size=arena_size)
+        self.setup_arena(size=arena_size)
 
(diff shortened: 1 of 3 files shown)


soc/hps_soc.py
```diff
@@ -152,11 +151,14 @@ def setup_ram(self, size):
         self.bus.add_slave("sram_lram", self.lram.bus, region)
         self.bus.add_region("sram", region)
 
+    # define the "arena" region even if it's length-zero
```
danc86, line 154: I guess we need this to avoid errors from the linker script referencing a non-existent region when the option is turned off, right?

danc86 (approved):
Thanks @tcal-x this looks great!

soc/hps_soc.py
tcal-x: Exactly.   I couldn't another way to do it (without doing something more complicated such as running the linker script through a preprocessor).

PR #343, opened 2021-11-18
Bump TFLite Micro

robertszczepanski:
This PR bumps TF Lite Micro and removes it form submodules. It has been generated with `create_tflm_tree.py` provided by TFLite Micro which is mentioned in #233.
This PR will be followed by a PR that adds sync CI which will automatically update `third_party/tflite-micro` nightly via PRs like in https://github.com/tensorflow/tflite-micro-arduino-examples. That PR will also add a way to update `tflite-micro` via `scripts/setup`. It is almost ready.

@tcal-x I have a question about `proj/mnv2_first/src/tensorflow/lite/kernels/internal/reference/sub.h`. How to properly update this file? Is it different than https://github.com/tensorflow/tflite-micro/blob/main/tensorflow/lite/kernels/internal/reference/sub.h?
Right now I just modified it manually so it seems to work with CFU Playground's demos but I don't think that's the expected solution.

PR #346, opened 2021-11-19
Add CI that automatically updates `third_party/tflite-micro`

robertszczepanski:
This PR should be merged after #343. It introduces new CI workflow that checks if there are any updates possible to apply on `third_party/tflite-micro`.

It also adds a script to update it manually. It is also included in `scripts/setup` so running a setup will also update `tflite-micro`. If there is already existing `third_party/tflite-micro` setup script will prompt a warning to not override an existing one. If user decides to remove it, then setup script will get the latest version.
Current TF Lite Micro version used by CFU Playground will be stored in `conf/tflite-micro.version`.

Since this was tested on Antmicro's fork, there is an Antmicro's bot account set up in CI. You can either let me know what should be a name and email of an account for CFU Playground or you can merge this PR and override it yourself. However you will need to set a secret token `GH_SERVICE_ACCOUNT_TOKEN` for this repo to make it work.

Here is how it looks: https://github.com/antmicro/CFU-Playground/pull/2

It is set up to enable auto-merge but I can also set it to assign someone as a reviewer on every PR created.

FYI @tcal-x

### DIFF
--- a/scripts/sync_from_tflite_micro.sh
+++ b/scripts/sync_from_tflite_micro.sh
@@ -1,0 +1,68 @@
+#!/usr/bin/env bash
+# Copyright 2021 The CFU-Playground Authors
+# Copyright 2021 The TensorFlow Authors. All Rights Reserved.
+#
+# Inspiried by https://github.com/tensorflow/tflite-micro-arduino-examples/blob/main/scripts/sync_from_tflite_micro.sh
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+set -ex
+
+SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+CFU_ROOT="$(dirname $(dirname $(realpath ${BASH_SOURCE[0]})))"
+cd "${ROOT_DIR}"
+
+TEMP_DIR=$(mktemp -d)
+cd "${TEMP_DIR}"
+
+echo Cloning tflite-micro repo to "${TEMP_DIR}"
+git clone --depth 1 --single-branch "https://github.com/tensorflow/tflite-micro.git"
+cd tflite-micro
+git rev-parse --short HEAD >| ${CFU_ROOT}/conf/tflite-micro.version
+
+make -f tensorflow/lite/micro/tools/make/Makefile clean_downloads
+
+OUTPUT_DIR="${TEMP_DIR}/tflm_tree"
+
+# Create the TFLM base tree, get person_detection for its models
+python3 tensorflow/lite/micro/tools/project_generation/create_tflm_tree.py "${OUTPUT_DIR}"
+
+# Download person_model_int8 third party. URL taken from tflite-micro:
+# https://github.com/tensorflow/tflite-micro/blob/main/tensorflow/lite/micro/tools/make/third_party_downloads.inc
+PERSON_MODEL_INT8_ZIP="tf_lite_micro_person_data_int8_grayscale_2020_12_1.zip"
+PERSON_MODEL_INT8_URL="https://storage.googleapis.com/download.tensorflow.org/data/${PERSON_MODEL_INT8_ZIP}"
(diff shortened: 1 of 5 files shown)


scripts/sync_from_tflite_micro.sh
```diff
@@ -0,0 +1,68 @@
...
+cp tensorflow/lite/micro/kernels/depthwise_conv_test* ${OUTPUT_DIR}/tensorflow/lite/micro/kernels
+cp -r tensorflow/lite/micro/kernels/testdata ${OUTPUT_DIR}/tensorflow/lite/micro/kernels/
+cp -r tensorflow/lite/micro/models ${OUTPUT_DIR}/tensorflow/lite/micro/
+cp -r tensorflow/lite/micro/tools ${OUTPUT_DIR}/tensorflow/lite/micro/
+cp -r person_model_int8/* ${OUTPUT_DIR}/tensorflow/lite/micro/examples/person_detection/
+
+# copy ${OUTPUT_DIR} to the repo
+cp -aT "${OUTPUT_DIR}" "${CFU_ROOT}/third_party/tflite-micro"
+
+rm -rf "${TEMP_DIR}"
```
tcal-x, line 68: Is there a reason that you remove `TEMP_DIR` but not `OUTPUT_DIR`?

tcal-x: Ah, I see, `OUTPUT_DIR` is a child of `TEMP_DIR`.

tcal-x (approved):
LGTM

tcal-x:
Hi @robertszczepanski , the runs are failing as expected (https://github.com/google/CFU-Playground/runs/4424822552?check_suite_focus=true).    So now it's time for the next step.

This is my first time using tokens so I'll need your help.   Do I go to the repo settings, "secrets", then create an environment?  You're not talking about a personal access token (https://github.com/settings/tokens), right?  

And for the name/e-mail, is that usually an e-mail alias/list of the repo admins?

mithro:
Please sync with @acomodi at Antmicro about how to do this with deploy keys.

PiotrZierhoffer:
Hi @tcal-x 

The flow we follow here is based on our work with TensorFlow Lite Micro (e.g. https://github.com/antmicro/tensorflow-arduino-examples). Looking back at it, it might be a little bit too complex for you, as it requires an additional user.

I will describe it below nevertheless. In few hours I will try to describe the way with SSH deploy keys, which does not require additional accounts and is generally easier to manage.

============ 

Secrets have to be defined here:
https://github.com/google/CFU-Playground/settings/secrets/actions

For tflite we created a separate account with write access. You should have a custom bot user as well. You could use your own account, but frankly speaking I would not recommend it.

The token is the Personal Access Token of that account and you need to create it manually.
You don't need (and should not) add too many permissions to this token. A single tick next to `public_repo` should be enough, but we can discuss it offline if it's not.

**Be aware that this token will not be readable after creation. You need to write it down somewhere.**

Then you need to change the username/mail to the account you created: https://github.com/google/CFU-Playground/pull/346/files#diff-099b8b414ddee5c6ab30545910ee726e5132ddbebbad8c2e2695596771186ba7R18

Last step is to create the secret variable `GH_SERVICE_ACCOUNT_TOKEN` with the proper value.

============ 

As I mentioned, in few hours I'll give you a description with fewer steps, that will be a bit easier to manage.

tcal-x:
@PiotrZierhoffer yes I would definitely be interested in the "deploy keys" approach if it is simpler, thanks!

PiotrZierhoffer:
Ok, after a few more iterations I believe we can follow the symbiflow-arch-defs example: https://github.com/SymbiFlow/symbiflow-arch-defs/blob/5f7c92ed02ebb1a5279a56e42e4d07bfc4bfe860/.github/workflows/update_conda_lock.yml

This will require no additional users.

Todo:

* with `ssh-keygen` create an SSH key pair. I recommend providing a distinctive name, as it will be used for this specific purpose only.
* in the deploy key settings add the content of the file with a public key generated in the previous step: https://github.com/google/CFU-Playground/settings/keys . Remember to check the "Allow write access" checkbox.
* add the content of the file with a private key as a secret variable: https://github.com/google/CFU-Playground/settings/secrets/actions . Name the variable `SSH_DEPLOY_KEY`.

With this we'll create a trivial PR that will adjust everything. 

Please note - this setup will NOT force CI on PR creation, only on push events. It's not really a big difference in this case, but it's important to keep all the workflows active both for PRs and pushes. I verified that right now everything is configured properly.

Please let me know if the above makes sense and ping me when you create these variable/deploy key.

tcal-x:
@PiotrZierhoffer , thank you!  

One question -- do I set a **passphrase** for the key pair?

mithro:
No

tcal-x:
@PiotrZierhoffer , I generated the key pair with:
```
ssh-keygen -t rsa -b 4096 -C "CFU Playground for TFLM"
```
and installed the keys as you described.

PR #347, opened 2021-11-20
Get and use SymbiFlow; includes CI test.

tcal-x:
To use (please suggest new names for the make targets if you don't like these):
* `make install-sf` -- takes about 6 minutes for me.  ~25GB of disk space.
* `make enter-sf` -- enter the environment with everything set up and in your path (including riscv gcc toolchain).  `exit` to exit.
* Add `USE_SYMBIFLOW=1` to your `make bitstream` line to specify SymbiFlow, or, if you're already using EXTRA_LITEX_ARGS, then instead add `--toolchain=symbiflow` to that.


alanvgreen (approved):
LGTM

PR #350, opened 2021-11-23
mnv2_first/src: remove unused customizations

alanvgreen:
Remove these ops. They were only customized to allow parameters to be
dumped to the console while exploring the model. This is no longer
required, and keeping these ops was complicating the TfLite uprev.


alanvgreen:
Unblocks #343

tcal-x (approved):
LGTM
Thanks Alan.

PR #358, opened 2021-11-30
Bump python deps to tip of main/master.

tcal-x:


tcal-x:
These are real failures, due to a changed memory allocation in Fomu and iCEBreaker.   The original 128kB "sram" has been split into "sram" and "main_ram", 64kB each.  I knew we'd need to make adjustments when I saw this change go in 2 months ago, but then I forgot about it.

alanvgreen (approved):
LGTM

PR #361, opened 2021-12-04
gen2/ram_input: New PixelAddressGenerator

alanvgreen:
Fetches input data from LRAM interface.


proj/hps_accel/gateware/gen2/ram_input.py
```diff
@@ -0,0 +1,398 @@
...
+
+The main mode is used when Conv2D input data has a depth that is a multiple of
+16 values per pixel. We call these groups of 16 pixels a "block". Each block
+begins in LRAM 0.
+
+In this main mode, four words are fetched concurrently, one each for four
+separate pixels, with one word being fetched from each LRAM. This produces four
+separate streams of pixel data. Assuming pixels are 16 bytes deep and thus
+consist of 1 block of data, the fetch order would be:
+
```
tcal-x, line 40: I'm still a bit unclear on the terminology.   Does the fetch from location 0 get 4 bytes, each belonging to a different pixel?   And 4 fetches x 4 bytes each means there are 16 pixels being processed concurrently in steady state?   Is a "block" 16 pixels * 16 depth = 256 bytes?

proj/hps_accel/gateware/gen2/ram_input.py
```diff
@@ -0,0 +1,398 @@
...
+|    0 |     0*  |     -   |     -   |     -   |
+|    1 |     1   |     4*  |     -   |     -   |
+|    2 |     2   |     5   |     8*  |     -   |
+|    3 |     3   |     6   |     9   |    12*  |
+|    4 |    16*  |     7   |    10   |    13   |
+|    5 |    17   |    20*  |    11   |    14   |
+|    6 |    18   |    21   |    24*  |    15   |
+|    7 |    19   |    22   |    25   |    28*  |
++------+---------+---------+---------+---------+
+
```
tcal-x, line 53: What is the purpose of the skew?    Does it make the summation more efficient?

proj/hps_accel/gateware/gen2/ram_input.py
alanvgreen: Oh, I messed up the comment here. Each group of 16 *bytes* is a block.

I've added another commit to clarify the wording, though I'm not sure how much it helps.

proj/hps_accel/gateware/gen2/ram_input.py
alanvgreen: There is a twofold purpose:
(a) since data for pixels all begin in the same LRAM bank (bank 0) we can't have four streams reading it at the same time. With this skew, there is no contention.
(b) the systolic array requires the input data be delivered with a skew. If it were not, we would need to introduce the delay.

tcal-x: Thanks that makes complete sense.  I understand now.

tcal-x (approved):
LGTM

danc86 (approved):

PR #362, opened 2021-12-06
Add CI to automatically update Renode version

robertszczepanski:
This is similiar to #346 but provides automatical updates for Renode version.

It updates Renode version in `conf/renode.version`.
`.github/workflows/test_project.yml` and `setup/scripts` are modified to get Renode version from that file. It is scheduled to run daily 2 hours after bumping TF Lite Micro so there shouldn't be any issues.

This CI would also require configuring a bot the same as in #346. It has been explained by @PiotrZierhoffer: https://github.com/google/CFU-Playground/pull/346#issuecomment-986848230

FYI @tcal-x


PiotrZierhoffer:
@tcal-x This is ready for review, but, as in the case of tflm, we need to decide on what email will be used for these bumps. It needs to be accepted by the Google CLA

PiotrZierhoffer:
Email corrected, ready for review.

tcal-x (approved):
LGTM

PR #363, opened 2021-12-06
Use a modified NXLRAM with dual ports and bank striping

tcal-x:
A modified version of litex/soc/cores/ram/lattice_nx.py
is added at ./soc/hps_lattice_nx.py:

 * optionally use dual port LRAM primitive
 * optionally stripe consecutive addresses across the banks

The dual port primitive and striping are enabled for the
separate tensor arena.

The second port is not yet made available to the CFU.


mithro (commented):
Some tests might be nice?

soc/hps_lattice_nx.py
```diff
@@ -0,0 +1,174 @@
...
+# This is a modified version of litex/soc/cores/ram/lattice_nx.py
+# Modified by CFU Playground Authors to:
+#  * optionally use dual port LRAM primitive
+#  * optionally stripe consecutive addresses across the banks
+##################################################################
+#
+#
+# This file is part of LiteX.
+#
+# Copyright (c) 2020 David Corrigan <davidcorrigan714@gmail.com>
```
mithro, line 11: Copyright should remain as first thing in the file.

mithro:
Also think this could benefit from some diagrams of how the blocks are laid out / connected.

soc/hps_lattice_nx.py
tcal-x: Fixed.

tcal-x: Also added CI tests with this option enabled (failure is masked, but that should get addressed soon, see #365 ).

Also added some ascii art in `hps_lattice_nx.py` showing the memory org.

mithro:
Don't wait for me to do more review, please get other people to review.

soc/hps_lattice_nx.py
```diff
@@ -0,0 +1,174 @@
...
+# Copyright (c) 2019 Tim 'mithro' Ansell <me@mith.ro>
+# Copyright (c) 2019 Florent Kermarrec <florent@enjoy-digital.fr>
+# SPDX-License-Identifier: BSD-2-Clause
+
+from migen import *
+from litex.soc.interconnect import wishbone
+
+kB = 1024
+
+"""
```
alanvgreen, line 48: For a string to be recognized as a  docstring, it must be the first statement of the module, class or function. See: https://www.python.org/dev/peps/pep-0257/#what-is-a-docstring

Two options:

1. Put the docstring right after the copyright comment, and have it include the expanded explanation you added, ncluding the table.
2. Use an explicit docstring with __doc__ = """\<stuff\>"""

I'd prefer (1)

alanvgreen (commented):
Looking good.

alanvgreen (approved):

soc/hps_lattice_nx.py
tcal-x: Thanks for pointing that out, updated!

PR #366, opened 2021-12-08
Set correct secrets for sync-tflm workflow

PiotrZierhoffer:


.github/workflows/sync-tflm.yml
```diff
@@ -31,12 +39,14 @@ jobs:
 
       - name: Sync the code, commit and push
         run: |
+          source environment
+          python3 -m pip install -r conf/requirements.txt
```
PiotrZierhoffer, line 43: @tcal-x I understand this might be a tiny bit controversial, but my reasoning behind this is: this is used only to be able to run a script that's contained within the CI. It's not required in regular builds, it does not affect users in any way.

We could use conda if it was used globally across the project, otherwise it's just making things complex for no real benefit.

PiotrZierhoffer:
@tcal-x  I'd consider this ready to be merged. The one question that remains is the email of the committer - it has to have Google CLA signed.

tcal-x:
Thanks @PiotrZierhoffer  -- I'm working on the e-mail right now.

PiotrZierhoffer:
Email corrected, ready for review.

tcal-x (approved):
LGTM

Issue #365, opened 2021-12-08
CI cleanup


tcal-x:
* We shouldn't need the SiFive RISC-V toolchain download any more
* We should stop masking failures in oxide runs -- this was from when the router would hang, leading to a timeout.   This has mostly been corrected.

tcal-x:
`fmax-trials-fresh-build.yml` will still need the SiFive download since it doesn't use Conda.

tcal-x:
In fact, having both the 8.3 SiFive toolchain and the litex-hub 10.1 toolchain causes link errors when building the software.

PR #367, opened 2021-12-09
Expose 2nd LRAM ports; connect to CFU; testcase.

tcal-x:

Update: I added a project (`proj/mport`) to test the new functionality.  It is tested in CI by the `oxide.yml` workflow.   In that flow I added doing the software compile, which encountered a clash of GCCs as expected (#365), so I removed the SiFive 8.3 toolchain download (there's still more to do for #365).

To enable all of the functionality added in this PR, you need to add both `--separate-arena` and `--cfu-mport` to your Litex flags.   This is already done in the Makefile in `proj/mport`.

### DIFF
--- a/soc/board_specific_workflows/general.py
+++ b/soc/board_specific_workflows/general.py
@@ -92,6 +92,11 @@ class GeneralSoCWorkflow():
         Returns:
             The LiteX Builder for the SoC.
         """
+        #for sp in soc.cpu.specials:
+        #    print("CPU special:", sp)
+        print("\n\nTJC\n\n")
+        print("cpu_params", soc.cpu.cpu_params)
+        print("cfu_params", soc.cpu.cfu_params)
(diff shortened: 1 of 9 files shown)


soc/board_specific_workflows/general.py
```diff
@@ -92,6 +92,11 @@ def build_soc(self, soc: litex_soc.LiteXSoC, **kwargs) -> builder.Builder:
         Returns:
             The LiteX Builder for the SoC.
         """
+        #for sp in soc.cpu.specials:
+        #    print("CPU special:", sp)
+        print("\n\nTJC\n\n")
```
danc86, line 97: May want to remove these before merging

tcal-x: Thanks for catching that!

danc86 (approved):
This looks good to me but I guess you may prefer to let Alan try it out before merging

alanvgreen (approved):

PR #373, opened 2021-12-10
Add a pause before connecting to the board. Helps Fomu.

tcal-x:
Fomu requires a second or so after the upload
for the tty to appear at /dev/ttyACM0.

Also add a "connect" target that does just that
(doesn't build anything).

### DIFF
--- a/proj/proj.mk
+++ b/proj/proj.mk
@@ -318,7 +318,7 @@ load: $(SOFTWARE_BIN)
 # Load hook allows common_soc.py to provide board-specific changes to load.
 # This isn't ideal, the logic is starting to get too voluminous for a Makefile.
 	$(SOC_MK) load_hook
-	@sleep 2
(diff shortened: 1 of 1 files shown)


proj/proj.mk
```diff
@@ -318,6 +318,11 @@ load: $(SOFTWARE_BIN)
 # Load hook allows common_soc.py to provide board-specific changes to load.
 # This isn't ideal, the logic is starting to get too voluminous for a Makefile.
 	$(SOC_MK) load_hook
+	@sleep 2
```
mithro, line 321: Don't use sleep like this.
Use something like;
```shell
while [ ! -f $(TTY) ]; do echo "Waiting for UART"; sleep 1; done
```

tcal-x: Thanks, works like a charm.

alanvgreen (approved):

PR #382, opened 2021-12-16
hps_accel/hps_cfu.h: add cfu_setx()

alanvgreen:
Adds cfu_setx() which had been omitted. Adding this allows the build to
succeed.


tcal-x (approved):
LGTM

PR #385, opened 2021-12-19
Update Verilator sim to handle dual-port arena

tcal-x:
Also fills in some missing sim support, e.g. handling custom VexRiscv variants.


soc/sim.py
```diff
@@ -79,6 +91,41 @@ def main():
...
+
+    if args.cfu_mport:
+        #
+        # add 4 read-only ports with LSBs fixed to 00, 01, 10, and 11.
+        #   and connect them to the CFU
+        #
+        newport = []
+        for i in range(4):
+            newport.append(soc.arena.mem.get_port())
+            soc.specials += newport[i]
```
alanvgreen, line 115: This is some kind of dark migen magic. Very neat.

alanvgreen (approved):
LGTM

PR #386, opened 2021-12-21
Use cache for Renode downloading in CI

PiotrZierhoffer:
This will save a lot of bandwidth and possibly speed up the whole CI process

I mark it as a draft, as it's still using Renode action from a branch. As soon as we merge it, I'll update this PR


PiotrZierhoffer:
The action is released, so this is ready to be reviewed. This allows us to download Renode once per a CI run instead of ~500 times.

tcal-x (approved):
LGTM

PR #391, opened 2021-12-23
Fix autogenerating Renode scripts

robertszczepanski:
This LiteX PR: https://github.com/enjoy-digital/litex/pull/1136 fixes Renode scripts autogeneration which makes it possible to generate HPS scripts.
Because of that I removed all predefined scripts since they didn't contain anything specific.

@tcal-x would you like to keep any of `.resc` scripts? I see contents of these are specific but since these are old and were used for `litex-vexriscv-tflite` target which does not exist anymore I removed them.

FYI @PiotrZierhoffer


tcal-x:
Thanks @robertszczepanski!  -- can you merge main into this branch (or rebase) to get Piotr's SymbiFlow update, so we can go green?

robertszczepanski:
@tcal-x -- I rebased it and tests are green now.

tcal-x (approved):
LGTM
Thanks!

Issue #380, opened 2021-12-15
ERROR: Module `\FD' referenced in module `\digilent_arty' in cell `\FD_7' is not part of the design.


tcal-x:
Caught in SymbiFlow CI tests.

This repo may be the only place where latest SymbiFlow is tested with (fairly) current LiteX.

The last successful SymbiFlow CI run in this repo was 2 days ago.

Here we **don't** pin the Conda package versions, so the tests here use the 'latest' packages.

tcal-x:
https://github.com/google/CFU-Playground/runs/4536982520?check_suite_focus=true

kgugala:
this is fixed in https://github.com/SymbiFlow/symbiflow-arch-defs/pull/2347

tcal-x:
Thanks @kgugala !

tcal-x:
FYI @PiotrZierhoffer

PiotrZierhoffer:
I believe this issue can be closed now?

PR #392, opened 2021-12-23
Bump the SymbiFlow toolchain to latest versions

PiotrZierhoffer:
This is to address #380


tcal-x (approved):
LGTM!
Thanks Piotr!

PiotrZierhoffer:
Credits to Karol ;) It turned out not to be related to yosys.

@tcal-x One question that arises here is if we want to stay on the latest version of the toolchain? Right now I set "current latest", but we can follow new releases if you want

PR #394, opened 2021-12-29
gen2i: Accelerate model with gen2

alanvgreen:
This PR allows gen2 to be used to accelerate all conv2D ops except the input layer.

For the conv2D layers that are accelerated, gen2 is about 10x the speed of gen1. Overall, gen2 currently operates at about half the speed of gen1.


danc86 (approved):
Wow, this is exciting! Does it imply that if we can also accelerate the input layer (layers? is there more than one input layer because of tiling?) then the overall performance should be 10x gen1?

alanvgreen:
> Wow, this is exciting! Does it imply that if we can also accelerate the input layer [...] then the overall performance should be 10x gen1?

I think gen2 will then be about 3x gen1. At that point, CONV2D will be about 20M cycles, and the biggest performance gains will come from steamlining the PAD (~40M cycles) and MAXPOOL (~20M cycles) operations.

> (layers? is there more than one input layer because of tiling?)

That's right. In the original model it's just one layer, but tiling splits it across several ops.

alanvgreen:
@tcal-x Not sure when you'll be able to take a look. Happy to revisit post-merge.

PR #396, opened 2022-01-03
gen2k: Implement input layer (model0) acceleration for hps_accel

alanvgreen:
Implements mode0 (input layer) acceleration for hps_accel (gen2).

After this PR, total execution time for the 09_20 model is 85M cycles, about 3x as fast as the gen1 accelerator overall (285M cycles). The Conv2D operations are approximately 10x as fast (19.2M cycles vs 217M).


proj/hps_accel/gateware/gen2/accelerator.py
```diff
@@ -57,7 +57,7 @@
     #  The maximum output value
     ('output_activation_max', signed(8)),
     # Address of start of input data, in 16 byte blocks (i.e addr 1 = byte 16)
```
danc86, line 62: I think this comment may also need changing? (now bytes instead of 16 byte blocks)

danc86 (approved):

proj/hps_accel/gateware/gen2/accelerator.py
alanvgreen: Done. Thank you.

PR #397, opened 2022-01-04
soc/hps_*.py: Fix slew rate and DDR clock

alanvgreen:
Fixes slew rate on IO pins issue to allow the SPI flash to reliably run
at full speed. Also uses DDR clocking on SPI flash. This reduces
HPS 09_20 evalation times by approximately 1M cycles.


danc86 (approved):

tcal-x (approved):
LGTM

PR #398, opened 2022-01-04
hps_accel/pad.h: Add unmodified pad.h

alanvgreen:
Accelerates the PAD operation for hps. This is a software-only change that reduces PAD times from 34M cycles to 3.3M cycles.


danc86 (approved):
Nice

PR #402, opened 2022-01-05
hps_accel/fully_connected: Accelerate fully connected

alanvgreen:
Accelerates fully_connected by a modest 10%, resulting in a 500k cycle speed up.


tcal-x (approved):
LGTM

danc86 (approved):
I suppose all gains from here onwards will be modest :-)

PR #404, opened 2022-01-06
hps/linker.ld: Fix large data address

alanvgreen:
Fixes large data in address space. This makes loading programs onto
proto2 faster since these data structures do not need to be updated if
they have not changed.

### DIFF
--- a/common/Makefile
+++ b/common/Makefile
@@ -44,7 +44,7 @@ GEN_LD_DIR := $(SOC_SOFTWARE_DIR)/include/generated
 LDSCRIPT   := $(LD_DIR)/linker.ld
 LDSCRIPTS  := $(LDSCRIPT) $(GEN_LD_DIR)/output_format.ld $(GEN_LD_DIR)/regions.ld
 
-SRC_DIR    := $(BUILD_DIR)/src
+SRC_DIR    := src
 
 PACKAGE    := software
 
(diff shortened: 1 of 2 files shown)


common/Makefile
```diff
@@ -44,7 +44,7 @@ GEN_LD_DIR := $(SOC_SOFTWARE_DIR)/include/generated
 LDSCRIPT   := $(LD_DIR)/linker.ld
 LDSCRIPTS  := $(LDSCRIPT) $(GEN_LD_DIR)/output_format.ld $(GEN_LD_DIR)/regions.ld
 
-SRC_DIR    := $(BUILD_DIR)/src
+SRC_DIR    := src
```
tcal-x, line 47: We are in `$(BUILD_DIR)`, right?   What's the motivation for this change?

tcal-x (approved):
LGTM

Does this PR  take care of everything we were trying to do in #323 , or is there some non-overlap?

common/Makefile
alanvgreen: Before this change, to identify a file by path, the linker script would need to include the absolute path name.

alanvgreen:
I had forgotten about #323 ! Thank you for pointing it out.

My motivation with this change is to shorten the load time during the development cycle - it currently takes 80s to load hps_accel software onto proto2. It might also bring some consistency to run times (depending on why they vary so much) but that is not the goal here.

tcal-x:
> My motivation with this change is to shorten the load time during the development cycle - it currently takes 80s to load hps_accel software onto proto2. 

I see, by having more fixed data at fixed locations,  the number of blocks of flash storage that don't change and therefore don't need to be rewritten is greatly increased.

alanvgreen:
Yes, that's exactly right. David has some neat MCU + host side code that
hashes SPI flash blocks and doesn't update on change, but we mostly haven't
taken advantage of it yet.

PR #416, opened 2022-01-12
Add custom nextpnr params

acomodi:
This PR adds the possibility to use a different build template that can be used to pass additional parameters to nextpnr

With some experimentation performed using [FPGA tool perf](https://github.com/SymbiFlow/fpga-tool-perf)'s ability to run a set of builds using different combinations of input parameters, one combination among others yielded good results (locally) both in terms of runtime and max frequency was the following:

```
placer-heap-beta: 0.5 (default)
placer-heap-alpha: 0.1 (default)
placer-heap-critexp: 0.7 (default)
placer-heap-timingweight: 52 (custom)
```

### DIFF
--- a/.github/workflows/fmax-trials-fresh-build.yml
+++ b/.github/workflows/fmax-trials-fresh-build.yml
@@ -16,7 +16,7 @@ jobs:
       - run: bash scripts/setup -ci
       - run: which pip3 && which python3 && which pip
       - run: riscv64-unknown-elf-gcc --version
-      - run: bash scripts/install_oxide 
+      - run: bash scripts/install_oxide
       - run: source environment && yosys --version && nextpnr-nexus --version
       - run: pwd && source environment && cd proj/proj_template_v && ulimit -S -t 900  && ulimit -H -t 900  && make PLATFORM=hps EXTRA_LITEX_ARGS="--cpu-variant=full     --nextpnr-seed=1"   clean bitstream |& egrep '(Max frequency|Info:.*%)' || true
       - run: pwd && source environment && cd proj/proj_template_v && ulimit -S -t 900  && ulimit -H -t 900  && make PLATFORM=hps EXTRA_LITEX_ARGS="--cpu-variant=full     --nextpnr-seed=22"  clean bitstream |& egrep '(Max frequency|Info:.*%)' || true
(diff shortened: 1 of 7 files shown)


tcal-x (approved):
LGTM!

tcal-x:
@acomodi -- can you double check the parameters you have above?   
Some of the defaults listed above don't match what I see in my `nextpnr-nexus`:
```
  --placer-heap-beta arg            placer heap beta value (float, default:           0.9)
  --placer-heap-alpha arg           placer heap alpha value (float, default:          0.1)
  --placer-heap-critexp arg         placer heap criticality exponent (int, default:   2)
  --placer-heap-timingweight arg    placer heap timing weight (int, default:          10)
```

Also, `placer-heap-critexp` has to be integer.

tcal-x:
Also, just changing `timingweight` did have some benefit but not the large benefit that was mentioned in the meeting.

acomodi:
@tcal-x For the parameters, I missed to clarify where those values come from. Basically the defaults can be overridden in two ways:

- command line arguments
- in-code assignments

In this case, the `beta` and `critexp` are statically assigned in the code and cannot be altered by the command line arguments equivalents: https://github.com/YosysHQ/nextpnr/blob/2ab08a872d08e6ba64a0b81303889a14941a88f0/nexus/arch.cc#L669-L670.

As for the `timingweight` effects, it may be not robust enough, but in CI the reached max frequency is the same as the one I see locally: https://github.com/google/CFU-Playground/runs/4810172300?check_suite_focus=true#step:14:23363

To be sure, what versions of the tools do you have?

Mine are:
```
Yosys 0.13+3 (git sha1 61324cf55, gcc 11.1.0-1ubuntu1~18.04.1 -fvisibility-inlines-hidden -fmessage-length=0 -march=nocona -mtune=haswell -ftree-vectorize -fPIC -fstack-protector-strong -fno-plt -O2 -ffunction-sections -fPIC -Os)
nextpnr-nexus -- Next Generation Place and Route (Version nextpnr-0.1-68-g2ab08a87)

```

tcal-x:
@acomodi  was this a typo above?

```
placer-heap-critexp: 0.7 (default)
```

acomodi:
@tcal-x Indeed, that should have been a `7` actually

tcal-x:
I did find one difference in my runs -- I was using `slim+cfu` (which is default in that project) instead of `slimopt+cfu`.   I'm rerunning with `slimopt+cfu` and `--placer-heap-timingweight 52` now.

tcal-x:
Using `slimopt+cfu` also helped a little -- I did crack 70MHz.

But I think I found the big difference -- in addition to `--separate-arena`, `--cfu-mport` must also be specified.   I'm a little puzzled why the build did not fail without this.

tcal-x:
@acomodi can you rerun your sweep adding `--cfu-mport` to see if you can find any potentially better params?

acomodi:
@tcal-x Sure, I have ran the sweep to get a better parameter, and found one to be `--placer-heap-timingweight 79` for the correct design that includes the `--cfu-mport`.

Unfortunately the design is quite huge and it is quite on the limit, in terms of size, to have more robust and stable results, but in general the above parameter seems to be giving good results. The latest run I had was gave me the following:

- Fmax: ~73 MHz
- Run-time for nextpnr-nexus (so P&R only): ~155 seconds

Adding the synthesis times which on avarage do not take more than 60 seconds we can have a build that works on HW and meets timing under the 4 minutes mark.

PR #422, opened 2022-01-18
Enable parallel software bitstream build

acomodi:
This PR enables the ability to parallely build the gateware and software so to reduce the overall build time.

It is tested on the final HPS design, that has been tested working on hardware.

**Note**: software compilation of the BIOS is disabled when building the gateware as it is not strictly needed. This is left to do only when running `make software`

Additionally, a `BUILD_FLAGS` variable is added to pass additional flags to build the software, such as the `-j` param to have also a parallel build of the software.


acomodi:
@tcal-x I have run CI only with the new target to see if it yields the same results as I get locally: https://github.com/google/CFU-Playground/runs/4854579081?check_suite_focus=true.

proj/hps_accel/Makefile
```diff
@@ -75,13 +75,19 @@ DEFINES += INCLUDE_MODEL_HPS
 # Use PLATFORM=common_soc TARGET=digilent_arty to select Arty instead of hps
 PLATFORM=hps
 
+# Parallel software and bitstream build
+PARALLEL ?= no
```
tcal-x, line 79: Having this variable `PARALLEL` is confusing since below we pass an option `--parallel` in `EXTRA_LITEX_ARGS` (which is the option to spawn `nproc` nextpnr jobs).   Let's use something different here like "PARALLEL_BUILD".

proj/hps_accel/Makefile
```diff
@@ -98,12 +104,19 @@ ifeq '$(TARGET)' 'digilent_arty'
 EXTRA_LITEX_ARGS += --sys-clk-freq 75000000
 endif
 
+ifneq '$(PARALLEL)' 'no'
+EXTRA_LITEX_ARGS += --extra-nextpnr-params
+EXTRA_NEXTPNR_ARGS += --placer-heap-timingweight 79
+endif
```
tcal-x, line 110: I feel these two things (parallel build & parameter adjustment) shouldn't necessarily be tied together.

tcal-x (changes requested):
See comments inline

proj/hps_accel/Makefile
tcal-x: Actually it's fine, we can adjust later as necessary.  However, is router1 used by default?   Do the params make sense with router1?

Unfortunately I have let it get quite complicated, between the options in Makefile and various scripts (and CI flows)...

proj/hps_accel/Makefile
acomodi: I have actually removed this variable as it is not required. I used it only to enable the extra nextpnr parameters

proj/hps_accel/Makefile
acomodi: I have removed this, as it is actually not in the scope of allowing parallel sw/hw builds

tcal-x (approved):
LGTM

PR #427, opened 2022-01-23
new models

alanvgreen:
Adds new models and provides a fix to the gateware to allow those models to work correctly. Includes the changes to nextpnr-nexus params as suggested by Antmicro.


danc86 (approved):

scripts/parallel-nextpnr-nexus
```diff
@@ -54,13 +55,14 @@ set -o pipefail
...
 mkdir runs
-seed_base=$((RANDOM % 100 * 100))
+seed_base=1000
 for s in $(seq ${SEED_COUNT}); do
   seed=$((seed_base + s))
   DIR="runs/seed-${seed}"
   mkdir "${DIR}"
   ( 
     nextpnr-nexus \
+        --placer-heap-timingweight 52 \
```
tcal-x, line 65: I think the new constant is 79, but we can do that in another PR.

tcal-x (approved):
LGTM

PR #429, opened 2022-01-24
Remove LEDChaser from hps_soc

keadwen:
They were nice to have in the early development stage.
Now they are messing up with my power measurements.


tcal-x (changes requested):
Hi Jakub, can we instead put it on a switch that's hardwired off, so that it's easy to turn back on if needed?

```
with_led_chaser = False

if with_led_chaser:
    ....
```

tcal-x:
Superseded by #439.

PR #431, opened 2022-01-24
Bump python dependencies (litex etc.).

tcal-x:


alanvgreen (approved):
LGTM

tcal-x:
I just did a build of this vs current main, and there was no change at all in either the bitstream or the software.   So this should be a no-op.

PR #433, opened 2022-01-25
[WIP] Use MULTADDSUB9X9WIDE for MaccBlock

mkurc-ant:
This PR adds manual instances of the `MULTADDSUB9X9WIDE` for `MaccBlock`. That macro maps directly to the module architecture, with it we can almost completely remove all generic LUTs and FFs currently needed.

The change requires nextpnr with this fix https://github.com/YosysHQ/nextpnr/pull/899 and results in a working design in hardware.

There are two issues here:
 - Tests for systolic array fail - nMigen/Amaranth cannot simulate behavior of the DSP macro instance.
 - If the design is to be targeted for multiple architectures there has to be conditional generation so that the macro is used only for Nexus.

### DIFF
--- a/proj/hps_accel/gateware/gen2/macc.py
+++ b/proj/hps_accel/gateware/gen2/macc.py
@@ -16,7 +16,7 @@
 """Multiply Accumulate Blocks for a systolic array"""
 
 
-from nmigen import Mux, Signal, unsigned
+from nmigen import Mux, Signal, unsigned, Const, Instance, ClockSignal, ResetSignal, Cat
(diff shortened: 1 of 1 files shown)


alanvgreen:
Regarding the issue of how to integrate so that tests still pass, I'll work up a PR to pass in a flag that allows for platform specific code.

I raised #434 to remind us to think about a better long-term solution.

alanvgreen:
Tested this PR and found it saved 1780 LUTs (wow!).

alanvgreen:
#437 wraps this code to address the two issues raised in the PR description (unit test failures, and lack of generality).

PR #435, opened 2022-01-25
parallel-nextpnr-nexus: better seed handling

alanvgreen:
Uses the incoming seed as the first seed rather then always starting
from 1000.


tcal-x (approved):
LGTM

tcal-x (approved):

PR #436, opened 2022-01-25
CI: sync-renode: Fix the up-to-date behavior

ajelinski:
Now the `git commit` really will only be executed when the new version
is available.

When Renode is up to date:
* the CI job will finish with a success status,
* the `create-pr` step won't start,
* a clear notice with such an information will be printed in the CI run.

Signed-off-by: Adam Jeliński <ajelinski@antmicro.com>


PiotrZierhoffer (approved):
Lgtm!

PR #437, opened 2022-01-25
Wraps: Manually instantiated MULTADDSUB9X9WIDE in MaccBlock to save resources.

alanvgreen:
Builds on #433 to ensure unit tests pass while allowing the use of the DSP macro when building for Crosslink NX.


tcal-x (approved):
LGTM

PR #441, opened 2022-01-27
hps: stop compiling Litex BIOS

danc86:
This makes it possible to drop the timer peripheral, and also generally
saves a small amount of time in builds.


danc86:
Sorry this doesn't actually work properly, I obviously did not test it adequately. Will fix it up.

danc86:
Commit 5c644452f0c324fbb3db490a9f2e07d299dc8d32 already did this

PR #443, opened 2022-01-28
Revert "Enable parallel software bitstream build"

danc86:
Reverts google/CFU-Playground#422

There are two issues with that commit:

* It introduces an unrelated change of the CPU variant from `slim+cfu` to 
  `slimopt+cfu`, which we may want to do, but needs to be considered separately 
  (the hps-firmware tree would need a corresponding update).

* There is no direct Make dependency from software compilation to Litex CSR 
  generation. There *is* an indirect dependency through the $(BIOS_BIN) target, 
  software/bios/bios.bin. But that target only depends on cfu.v, but not other 
  Litex generated outputs.

  It means that a change to the Litex SoC definition could change CSRs but the 
  software would be wrongly compiled against the previous CSR definitions, 
  leading to confusing runtime failures.

Building the software concurrently with the bitstream is also not a very useful 
improvement. Although it may help CI somewhat, for typical developer workflows 
the software compilation takes negligible time compared to the bitstream 
synthesis.


danc86:
Amended to add a proper commit message, instead of the Github-generated one.

danc86:
Another problem I noticed is that we actually can't pass `--no-compile-software` either. Although we don't ever use the Litex BIOS on HPS, we still need to let Litex compile its libbase because we link against that.

tcal-x (approved):

PR #448, opened 2022-02-02
hps: allow removing timer by not building BIOS code

acomodi:
This PR does the following:

- adds an additional argument for the HPS SoC to control whether the BIOS needs to be built. The default behavior is to not build the BIOS, therefore removing the unnecessary `timer0`, saving up a couple hundreds of LUTs and FFs.
- Adds the `--no-compile-software` flag when building the bitstream. CSRs are still being output, and they are also relevant only when building software. Given that changes either to the gateware or software in terms of CSR requires building both of them, it should be alwasy the case to first do `make clean` and than `make bitstream software`. 
- Adds possibility to build the HPS software parallely by adding the `BUILD_JOBS=<n jobs>` when running `make software`

Signed-off-by: Alessandro Comodi <acomodi@antmicro.com>

### DIFF
--- a/proj/hps_accel/Makefile
+++ b/proj/hps_accel/Makefile
@@ -102,9 +102,16 @@ ifeq '$(TARGET)' 'digilent_arty'
 EXTRA_LITEX_ARGS += --sys-clk-freq 75000000
 endif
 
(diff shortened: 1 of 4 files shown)


tcal-x (approved):
LGTM

PR #459, opened 2022-02-08
Update Setup Guide for Arty A7 using Symbiflow

APIUM:
Add note about symbiflow route and place tools not completing if
unable to meet timing. This will be likely to occur using the
suggested Arty A7 when completing the Test Run in Step 6 of the
Setup Guide. As this can be confusing for new users, this change
suggests the user can add the command to lower the target clock
rate when completing this step if required.

### DIFF
--- a/docs/source/setup-guide.rst
+++ b/docs/source/setup-guide.rst
@@ -79,6 +79,12 @@ This starts a subshell; you can leave the environment by typing ``exit``.
 Later when you're building a project bitstream, 
 you need to add USE\_SYMBIFLOW=1 to your make command line.
 
+If you see no errors, but the compiled RISC-V executable does not run on the
+board, you may need to lower the target clock rate.
+This can be done by adding ``EXTRA_LITEX_ARGS="--sys-clk-freq {clockrate}"``
+to the make command, where {clockrate} if the desired clock rate in Hz.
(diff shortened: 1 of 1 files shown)


docs/source/setup-guide.rst
```diff
@@ -79,6 +79,11 @@ This starts a subshell; you can leave the environment by typing ``exit``.
 Later when you're building a project bitstream, 
 you need to add USE\_SYMBIFLOW=1 to your make command line.
 
+If your command does not succeed you may need to lower the target clock rate.
```
tcal-x, line 82: I think we need to be a little bit more precise here, since the confusing thing is that the bitstream generation did not give a clear error, so you didn't see any problem until you tried to run the software on the bitstream on the board.  Maybe "If you see no errors, but the compiled RISC-V executable does not run on the board, you may need to lower the target clock rate" -- sound ok?

tcal-x (changes requested):

docs/source/setup-guide.rst
APIUM: Thanks for the suggestion, I agree - have pushed an update.

tcal-x (approved):
LGTM

tcal-x:
Thanks @APIUM , this is merged.

PR #463, opened 2022-02-09
scripts/parallel-nextpnr-nexus: error code

alanvgreen:
Exit with an error code if place and route did not succeed.


tcal-x (approved):
LGTM

PR #465, opened 2022-02-10
hps: reduce OSCA worst case to 7%

danc86:
The current version of Crosslink-NX Family Data Sheet lists the high
frequency oscillator maximum frequency as 481.5MHz (that is, 7% higher
than its nominal 450MHz):

https://www.latticesemi.com/-/media/LatticeSemi/Documents/DataSheets/CrossLink/FPGA-DS-02049-1-2-1-CrossLink-NX-Family.ashx?document_id=52780

Older documents listed a wider frequency range but ±7% is the range for
production parts.

### DIFF
--- a/soc/hps_proto2_platform.py
+++ b/soc/hps_proto2_platform.py
@@ -70,11 +70,11 @@ class _CRG(Module):
         # Clock from HFOSC
         self.submodules.sys_clk = sys_osc = NXOSCA()
         sys_osc.create_hf_clk(self.cd_sys, sys_clk_freq)
-        # We make the period constraint 10% tighter than our actual system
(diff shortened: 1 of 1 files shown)


tcal-x (approved):
LGTM

tcal-x:
@danc86 , we need both this _and_ the [change to Nextpnr](https://github.com/YosysHQ/nextpnr/pull/912/files)?   We're not over-correcting are we?

danc86:
The nextpnr constraint is only a default, which is used if the design doesn't have a constraint for the same clock in its pdc. In the hps_*_platform.py files have always specified an explicit constraint so the nextpnr default one is irrelevant for us.

PR #471, opened 2022-02-14
Vizier scripts for HPS platform

kboronski-ant:
Integration with Google Vizier platform for HPS gen2.

Simply fill the required fields in `proj/hps_accel/vizier_autotune.ipynb` and run the notebook.
You can change the parameters to be adjusted and metrics to be optimized by modyfying the study definition structure. Parameter names are indentical as names used by `nextpnr-nexus`. See the notebook for more details.

The most universal way of running this notebook (headless, requires no DE, except for Google authentication which is done through browser. To avoid that you need to provide cached credentials for `gcloud`):
```
cd proj/hps_accel
jupyter nbconvert --to notebook --execute vizier_autotune.ipynb --output=output.ipynb
```

Using jupyter-lab may cause severe performance issues when displaying synthesis output. Classic jupyter notebook and Jupter extension for VS Code handle the output fine.

You will be asked to authenticate using your Google account twice.
That's just how Google handles the authentication process.

### DIFF
--- a/proj/hps_accel/vizier/client_vizier.py
+++ b/proj/hps_accel/vizier/client_vizier.py
@@ -1,0 +1,188 @@
+import math
+import random
+import time
+
+from google.cloud import storage
+from googleapiclient import discovery, errors
+from scipy.fftpack import idct
+
+from vizier.client import OptimizerClient
+from vizier.vapi import *
+
+_OPTIMIZER_API_DOCUMENT_BUCKET = 'caip-optimizer-public'
+_OPTIMIZER_API_DOCUMENT_FILE = 'api/ml_public_google_rest_v1.json'
+
+random.seed()
+
+def _read_api_document(project_id):
+    client = storage.Client(project_id)
+    bucket = client.get_bucket(_OPTIMIZER_API_DOCUMENT_BUCKET)
+    blob = bucket.get_blob(_OPTIMIZER_API_DOCUMENT_FILE)
+    return blob.download_as_string()
+
+def _trial_name(project_id, region, study_id, trial_id):
+  return f'projects/{project_id}/locations/{region}/studies/{study_id}/trials/{trial_id}'
+
+def trial_from_gtrial(study: Study, id, gtrial):    
+    trial = Trial(
+        client_id = id,
+        state = gtrial['state'],
+        parameters = [],
+        measurements = []
+    )
+
+    for param in gtrial['parameters']:
+        pname = param['parameter']
+        ptype = study.parameters[pname].type
+        pval = None
+        if ptype in ['DOUBLE', 'DISCRETE']:
+            pval = param['floatValue']
+        elif ptype == 'INTEGER':
+            pval = param['intValue']
+        elif ptype == 'CATEGORICAL':
+            pval = param['stringValue']
+        else:
+            raise Exception('Unknown parameter type')
+
+        p: Parameter = study.parameters[pname].trial(pname, pval)
+        trial.parameters += [p]
+    
+    return trial
+
+class VizierClient(OptimizerClient):
+    def __init__(self, account, project_id, region):
+        self.account = account
+        self.project_id = project_id
+        self.region = region
+        self.client_id = f'client{"".join([str(random.choice([0, 1, 2, 3, 4, 5, 6, 7, 8, 9])) for i in range(0, 8)])}'
+        self.api = discovery.build_from_document(service=_read_api_document(self.project_id))
+    
+    def create_study(self, study: Study, concurrency: int):
+        # study_id = f'{self.account}_study_{datetime.datetime.now().strftime("%Y%m%d_%H%M%S")}'
+        req = self.api.projects().locations().studies().create(
+            parent=f'projects/{self.project_id}/locations/{self.region}',
+            studyId=study.name,
+            body={'study_config': study.to_json_dict()}
+        )
+        try:
+            print(req.execute())
+        except errors.HttpError as e:
+            if e.resp.status == 409:
+                print('Study already existed.')
+            else:
+                raise e
+    
+    def get_suggestions(self, study: Study, concurrency: int) -> 'list[Trial]':
+        resp = self.api.projects().locations().studies().trials().suggest(
+            parent = f'projects/{self.project_id}/locations/{self.region}/studies/{study.name}',
+            body = { 'client_id': self.client_id, 'suggestion_count': concurrency }
+        ).execute()
+        op_id = resp['name'].split('/')[-1]
+        get_op = self.api.projects().locations().operations().get(
+            name = f'projects/{self.project_id}/locations/{self.region}/operations/{op_id}'
+        )
+
+        # Wait for GCloud to be done
+        while True:
+            op = get_op.execute()
+            if 'done' in op and op['done']:
+                break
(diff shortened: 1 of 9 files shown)


tmichalak:
@proppy

proj/hps_accel/vizier_autotune.ipynb
```diff
@@ -0,0 +1,295 @@
+{
```
proppy, line 1: you may want to consider https://jupytext.readthedocs.io/, if you want to maintain this as plain python script.

proj/hps_accel/vizier/client_vizier.py
```diff
@@ -0,0 +1,188 @@
...
+            pval = param['stringValue']
+        else:
+            raise Exception('Unknown parameter type')
+
+        p: Parameter = study.parameters[pname].trial(pname, pval)
+        trial.parameters += [p]
+    
+    return trial
+
+class VizierClient(OptimizerClient):
```
proppy, line 52: did you consider using https://googleapis.dev/python/aiplatform/latest/aiplatform_v1beta1/vizier_service.html? that might allow you to get rid of the helper classes.

proj/hps_accel/vizier_autotune.ipynb
```diff
@@ -0,0 +1,295 @@
...
+    "    max_trials = 3,\n",
+    "    # Add metrics to optimize here\n",
+    "    metrics = {\n",
+    "        'fmax': 'MAXIMIZE',\n",
+    "        'runtime': 'MINIMIZE'\n",
+    "    },\n",
+    "    # Add parameters for the blackbox here.\n",
+    "    # Use 'UNSPECIFIED'/'LENEAR'/'LOG'/'REVERSE_LOG' to define parameter scaling (Vizier)\n",
+    "    parameters = {\n",
+    "        'estimate-delay-mult': ParameterInt(min=10, max=50, scale='LINEAR')\n",
```
proppy, line 121: consider using https://papermill.readthedocs.io/en/latest/usage-parameterize.html pattern to parametize the notebook, so that it can easily be batched using: https://cloud.google.com/vertex-ai/docs/workbench/managed/schedule-managed-notebooks-run

proj/hps_accel/vizier_autotune.ipynb
```diff
@@ -0,0 +1,295 @@
...
+    "###  Install python dependencies"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "! pip install -U google-api-python-client\n",
```
proppy, line 25: nit: curious if the space suffix a convention you're following from another project? (I'm personally always have seen those without an extra space)

proj/proj.mk
```diff
@@ -298,6 +298,9 @@ prog: $(CFU_VERILOG)
 bitstream: $(CFU_VERILOG)
 	$(SOC_MK) bitstream
 
+synth: $(CFU_VERILOG)
```
proppy, line 301: does this also need to be implemented for non-hps platforms?

proppy (commented):
Thanks for sharing! An initial round of cosmetic comments, I didn't already try it yet :)

proj/hps_accel/vizier/client_vizier.py
kboronski-ant: I've mostly followed the examples from here, which do not use that: https://cloud.google.com/ai-platform/optimizer/docs

The `OptimizerClient` class that `VizierClient` is built on top is an abstraction layer that allows other back-ends, such as Advisor, to be used in place of Vizier, but implementations of their respective clients is not included within this PR.

proj/hps_accel/vizier_autotune.ipynb
proppy: I see :) thanks for the reference, I'm not sure if I personally like it but at least it's consistent :)

proj/hps_accel/vizier/client_vizier.py
proppy: I see, it's nice that the implementation is swappable but I don't think that prevent us from reusing existing high level client to implemented the plumbing with vizier.

I also found https://www.tensorflow.org/cloud/guides/cloud_tuner_guide which seems to provide an even higher-level interface (and is also used in the official samples: https://cloud.google.com/ai-platform/optimizer/docs/optimizing-multiple-objectives)

proj/proj.mk
proppy: @tcal would the regular platform (non-hps) also benefit from this?

tcal-x: This particular line (301) (the 'synth' target) is available for all non-sim builds, including regular (common_soc, non-hps) platforms like Arty, Fomu, and Icebreaker.   

If you mean, should we make the entire Vizier framework usable for these other platforms -- it depends on the parameterizations possible with other toolschains.   Yosys has a few options, but they're typically Boolean (on/off), not a numeric parameter.   I don't know offhand if there are similar parameters that we have here with nextpnr-nexus with the other PnR tools -- nextpnr-ice40, nextpnr-ecp5, vpr, or Vivado.  So I guess for each tool, we'd want a table of parameters that can be searched.

There are many parameters to the VexRiscv that could be searched.   @proppy , do you think it would make sense to do that search separately, or combined with the toolchain parameter search?

proppy: @tcal-x I'd be tempted to have one notebook per parameter "kind" we want to explore (one around pnr, another around risc-v features) so that we can strongly correlate the metrics that we want to score with the parameters we're trying to changes but it'd be also interesting to see how well vizier perform with concurrent parameters exploration.

proj/hps_accel/vizier/client_vizier.py
proppy: There might also be an deeper way to integrate this into the Vertex AI platform using: https://cloud.google.com/vertex-ai/docs/training/using-hyperparameter-tuning

They basically have a high level [method](https://googleapis.dev/python/aiplatform/latest/aiplatform_v1/job_service.html?highlight=create_hyperparameter_tuning_job#google.cloud.aiplatform_v1.services.job_service.JobServiceClient.create_hyperparameter_tuning_job) that allow to create a full hyperparameter search pipeline and point it to a python entrypoint that takes the parameters as command line arguments and report metrics by writing them to disk, see https://cloud.google.com/vertex-ai/docs/training/code-requirements#write_code_for_hyperparameter_tuning

The nice thing is that it allow you to visualize the progress the Vertex AI UI:
image

I'm not sure how partical it is for non-AI workflow, but I'd be happy to give it a try if you think that helps.

tcal-x:
Hi @kboronski-ant , can you sign off on your commits?   There are instructions here: https://github.com/google/CFU-Playground/pull/471/checks?check_run_id=5189400848

tmichalak:
@proppy @tcal-x I believe you can review, so that this can be eventually merged.

proppy (approved):

proppy:
Happy to handle the changes suggested during review as a followup PR.

tcal-x:
@kboronski-ant @tmichalak can you remove the [WIP] from the title to make it clear that it's ready for merge?

tcal-x (approved):

PR #473, opened 2022-02-16
Filter the sync workflows to only run on the main repo

PiotrZierhoffer:
It turned out that detecting if you are a fork in runtime is hell, but there's a convenient check we might have. It has a hardcoded "google" string, so please let me know if it's ok


tcal-x:
The two failures were due to apt install failures, so I'm ignoring them and merging.

PR #475, opened 2022-02-16
Bump Lite* including litespi to get CSR area reduction.

tcal-x:


danc86 (approved):
Thanks @tcal-x. There's no area reduction from Litespi unless we also pass with_csr=False, but we can do that now after this bump.

Issue #455, opened 2022-02-06
Unable to load the RISCV program (Setup Guide) onto the board when using Symbiflow


APIUM:
Board: ARTY A7
OS: Ubuntu 20.04 VM
TERM: Tested both `urxvt` and `gnome-terminal`
Running latest master of CFU-Playground, following [Setup Guide](https://cfu-playground.readthedocs.io/en/latest/setup-guide.html)

----

Command `make load USE_SYMBIFLOW=1` does not finish, stuck at `/home/{user}/projects/CFU-Playground/soc/bin/litex_term --speed 1843200  --kernel /home/{user}/projects/CFU-Playground/proj/proj_template/build/software.bin /dev/ttyUSB1`.

Pressing ENTER or typing `ENTER` as the Setup Guide states does not work, have left it stuck at that line for several hours with no change. The previous step `make prog` completes successfully with the LED sequence displayed on the ARTY. 

Rerunning script/setup has no change, all expected packages are installed. 

Works with Vivado with command `make load` - Golden tests pass successfully. I don't think this is an issue with the USB passthrough as a result, but not sure what else it could be. Has this been observed before?

tcal-x:
@APIUM , thank you for the report!

I think this is due to the place and route tools not meeting timing, but not reporting on it.  So the generated bitstream just won't work at 100MHz.   I'm working to address that (https://github.com/verilog-to-routing/vtr-verilog-to-routing/issues/1928).

In the mean time, we can try reducing the target clock rate for the Symbiflow build.    Let me verify that I have the correct incantation to do that, then I'll post the instructions here, then you can check if it works for you.

tcal-x:
@APIUM , this worked for me on my Arty board.   That's not a 100% guarantee it will work on your board.   This drops the system clock frequency to 75MHz:
```
% make prog USE_SYMBIFLOW=1 EXTRA_LITEX_ARGS="--sys-clk-freq 75000000"
% make load USE_SYMBIFLOW=1 EXTRA_LITEX_ARGS="--sys-clk-freq 75000000"
```

Note that you can't drop the frequency arbitrarily low; when you get close to 60MHz or lower, the DDR3 memory will not function correctly.

APIUM:
Thanks @tcal-x that worked perfectly for me. I've updated the documentation to suggest that users following the Setup Guide with the recommended hardware try this: https://github.com/google/CFU-Playground/pull/459.

APIUM:
[...]

PR #483, opened 2022-02-18
Make Arty default clk freq 75MHz when using SymbiFlow.

tcal-x:
This PR changes the default target frequency for the Arty A7 board to 75MHz when SymbiFlow is used.   It can still be overridden by an explicit `--sys-clk-freq` LiteX option to be either higher or lower.

The default frequency for an Arty A7 board in litex-boards is 100MHz.  When the bitstream is built with this target frequency, vpr will not meet timing, but it will not give visible error, and a bitstream will be generated.   The bitstream will not work on the board; the LEDChaser works, but loading and running software over UART does not work (see #455).

There is an open issue for vpr to produce a hard error when it does not meet timing (https://github.com/verilog-to-routing/vtr-verilog-to-routing/issues/1928).   However, here we would need to downgrade the error, since with even a simple CFU, the achieved maximum frequency (fmax) is 50MHz, but we can't drop the requested frequency that low, since at that frequency DDR memory does not function.

Experiments show that building for 75MHz and running on the board at that frequency usually works, even though vpr reports an fmax of about 50MHz.   Whether it actually works on a given board is dependent on the particular device though, so it is not guaranteed to work in all cases.

This current workaround would be better if there was a very visible message visible to the user along the lines of "75.0MHz requested but only xx.xMHz achieved.   A bitstream has been produced, but it is not guaranteed to work at 75.0MHz."

Followup questions:
* Is vpr being too conservative in its fmax calculation?   Running the fasm through fasm2bels --> Vivado timing analysis could give us some useful information.
* Is there a plan to improve fmax so that we can target 100MHz and meet timing with high confidence?   Is the issue that DSP blocks are not used?

### DIFF
--- a/soc/board_specific_workflows/digilent_arty.py
+++ b/soc/board_specific_workflows/digilent_arty.py
@@ -43,6 +43,8 @@ class DigilentArtySoCWorkflow(general.GeneralSoCWorkflow):
                             help='Arty variant: a7-35 (default) or a7-100')
         arty_args = parser.parse_known_args()[0]
         args = argparse.Namespace(**vars(arty_args), **vars(args))
+        if args.toolchain == 'symbiflow' and not args.sys_clk_freq:
+            args.sys_clk_freq = 75000000
         super().__init__(args, soc_constructor, builder_constructor)
 
     def make_soc(self, **kwargs) -> litex_soc.LiteXSoC:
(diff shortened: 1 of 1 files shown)


mithro (commented):
A comment explaining why the `sys_clk_freq` is being changed (and when this workaround can be removed) would be good.

tcal-x:
> A comment explaining why the `sys_clk_freq` is being changed (and when this workaround can be removed) would be good.

Yeah, this definitely needs that; added to the description.

mithro:
> > A comment explaining why the `sys_clk_freq` is being changed (and when this workaround can be removed) would be good.
> 
> Yeah, this definitely needs that; added to the description.

I actually meant in the code as a person is unlikely to see the commit message.

tcal-x:
> I actually meant in the code as a person is unlikely to see the commit message.

And they're even less likely to see the message in the PR :)     Comment has been added to the script.

tcal-x:
@ShvetankPrakash PTAL

ShvetankPrakash (approved):

PR #485, opened 2022-02-19
Use LiteX-generated $CPUFLAGS instead of hardcoded flags.

tcal-x:
The hardcoded "-march=rv32im" would be wrong if we try
using compressed or floating point extensions.


mithro (approved):
LGTM!

PR #490, opened 2022-02-25
Use the DSP FF Yosys plugin for HPS

mkurc-ant:
This PR enables the use of DSP register inference Yosys plugin by changing the synthesis script generated from Litex/Migen.

### DIFF
--- a/scripts/setup
+++ b/scripts/setup
@@ -128,8 +128,8 @@ if ! which yosys >/dev/null || ! yosys -V | grep "Yosys 0.1" >/dev/null; then
     if [ ! -e "${CFU_ROOT}/third_party/usr/local/bin/yosys" ]; then
         echo "Getting Yosys."
         mkdir -p "${CFU_ROOT}/third_party/usr/local"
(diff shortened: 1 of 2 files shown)


robertszczepanski:
Hi @tcal-x, it looks like test fail because there are 2 instances of yosys in an environment right now. One is installed by `make env` and it's in conda environment. The other one is installed by `scripts/setup` and it's in `third_party/usr/local`.

Is it necessary to keep the second one? Maybe we could add conda environment to readme and use it as a default one. The other benefit of that is also having a riscv toolchain already present in conda. CFU-Playground's configuration would be simplified.

If you want to keep both, I guess we can add a switch similiar to [this one](https://github.com/google/CFU-Playground/blob/main/scripts/setup#L62) and Yosys installation would be omitted in CI.`

tcal-x:
@robertszczepanski , I bumped the local yosys / yosys-symbiflow-plugins to get the dsp-ff plugin.   All tests pass now, but please double check that the new DSP tests give the results you expect.

I also rebased on current main.

robertszczepanski:
@tcal-x It's been tested on hardware and golden tests pass so it works as expected.

tcal-x (approved):

PR #504, opened 2022-03-21
Make the 'expects' script more robust.

tcal-x:
Tested with icebreaker, arty, and renode ('make run-renode').


ShvetankPrakash (approved):
Looks good to me.

PR #509, opened 2022-03-23
Expand space for .text section from 0x80000 to 0x81000 on HPS platform

tcal-x:
HPS tests were failing with the latest tflite-micro bump (https://github.com/google/CFU-Playground/pull/500#issuecomment-1075741023).   With local testing, I observed the same thing also on `main` branch, which is a bit puzzling.   This is the error (the exact value can vary):
```
/home/runner/work/CFU-Playground/CFU-Playground/proj/example_cfu/build/ld/linker.ld:27
cannot move location counter backwards (from 00000000202803f0 to 0000000020280000)
```

To reproduce the previous behavior locally,
```
cd proj/example_cfu
make PLATFORM=hps BUILD_JOBS=8 software
```
Edit: Another mystery -- the symptom only occurs eith `example_cfu` -- not with `proj_accel_1`, `mnv2_first`, `hps_accel`, or the  template projects.    I haven't found what the difference is.

Edit#2: Solved!   All of the other projects use `DEFINES += NDEBUG` to remove debug code.   When I add that to `example_cfu`, it also passes.    @alanvgreen , would you prefer that we go ahead and expand the .text section anyway, or, just add NDEBUG to the failing test?    Edit: why not both?  I've pushed another commit.

### DIFF
--- a/common/_hps/hps/ld/linker.ld
+++ b/common/_hps/hps/ld/linker.ld
@@ -24,7 +24,7 @@ SECTIONS
 		/*
 		 * Align models and large chunks of data into a constant position
 		 */
-		. = _ftext + 0x80000;
+		. = _ftext + 0x81000;
 		_frodata = .;
 		src/models/*/*(SORT(.rodata.*))
 		src/conv2d_??.o(SORT(.rodata.*))  /* specific to hps_accel */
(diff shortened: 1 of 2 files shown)


alanvgreen:
Nicely found!

tcal-x:
@alanvgreen PTAL  (I added some more notes to the description at the top)

ShvetankPrakash (approved):
Sorry was not able to approve this earlier but nice eye to catch that bug!

tcal-x:
@ShvetankPrakash I should go over the background story of this linker.ld file with you the next time we chat.

PR #511, opened 2022-03-24
Initial code for building customized VexRiscv on-demand.

tcal-x:
Prototype code for generating a VexRiscv variant on demand.   To trigger, just use any unrecognized variant, for example, add this to the project Makefile:

```
export EXTRA_LITEX_ARGS
EXTRA_LITEX_ARGS += --cpu-variant doesnotexist+cfu
```

The current code ignores the variant and just builds a vanilla VexRiscv.   More code needs to be added to parse the variant string to generate the config and the corresponding canonical Verilog file name.

### DIFF
--- a/soc/patch_cpu_variant.py
+++ b/soc/patch_cpu_variant.py
@@ -102,8 +102,24 @@ def build_cpu_variant_if_needed(variant):
     if variant in core.CPU_VARIANTS:
         print(f"Variant \"{variant}\" already known.")
     else:
+        #
+        #  TODO: parse "variant" to determine configuration in "gen_args",
+        #    and create a canonical "cpu_filename_base" that is uniquely
+        #    determined by the config.
+        #
+        dsz = 4096
+        isz = 4096
+        gen_args = []
+        if "cfu" in variant:
+            gen_args.append(f"--cfu=true")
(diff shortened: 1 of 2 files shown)


ShvetankPrakash:
Does this code change also patch it into the cpu [variants](https://github.com/google/CFU-Playground/blob/33f001087597d4168ed835fce7baa071e9fa0977/soc/patch_cpu_variant.py#L29) dictionary? Assuming you tested this change to make sure it did?

soc/patch_cpu_variant.py
```diff
@@ -93,3 +93,67 @@ def copy_cpu_variant_if_needed(variant):
...
+
+        #
+        # do some patching
+        #  TODO: 'arch' will depend on the CPU config.
+        #
+        arch  = "rv32im"
+        abi   = "ilp32"
+        gcc_flags = f"-march={arch} -mabi={abi}"
+
+        core.CPU_VARIANTS.update({
```
ShvetankPrakash, line 149: Ah nevermind I see it adds it here!

ShvetankPrakash:
Okay looks good to me. We can discuss the TODOs as next action item.

ShvetankPrakash (approved):

PR #514, opened 2022-03-29
CFU clock gating

mkurc-ant:
This PR adds a clock gate for the HPS CFU controlled via a CSR accessible from the risc-v core. The CSR has two bits: bit0 for clock enable and bit1 for reset assertion. This allows to turn the clock on/off from the software level as well as resetting the CFU gateware. There is a menu added that allows manually controlling this two bits.

The clock gate is built using the CrossLink-NX `DCC` primitive hence it will only work on the `nexus` arch for now.

I took some current measurements on the `hps_proto2` board using built-in shunt resistors for cases when the CFU clock was enabled and disabled. Turning the clock off decreases power consumption by about **6 to 10 mW**

| Rail | Voltage [V] | Clock on [mV] | Clock off [mV] | Rsense [ohm] | Clock on [mA] | Clock off [mA] |
| --- | --- | --- | --- | --- | --- | --- |
|FPGA_VCC1/2/3 |1.0 |4.5 |3.9 |0.1 |45 |**39** |
|Whole board |3.3 |2.3 |2.0 |0.1 |23 |**20** |

This is meant to be a proof of concept, not a final solution.

### DIFF
--- a/common/src/cfu_clk_menu.c
+++ b/common/src/cfu_clk_menu.c
@@ -23,12 +23,26 @@
 
 static void do_cfu_clk_enable () {
     puts("Enabling CFU clock\n");
-    cfu_ctl_csr_write(1);
+    uint32_t csr = cfu_ctl_csr_read();
+    cfu_ctl_csr_write(csr |  0x01);
 }
 
 static void do_cfu_clk_disable () {
     puts("Disabling CFU clock\n");
-    cfu_ctl_csr_write(0);
+    uint32_t csr = cfu_ctl_csr_read();
+    cfu_ctl_csr_write(csr & ~0x01);
+}
+
+static void do_cfu_rst_enable () {
+    puts("Asserting CFU reset\n");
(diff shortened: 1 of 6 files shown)


mkurc-ant:
I've done an additional test where I added a functionality to turn off the clock completely from the software. I added additional CSR bit that controls `OSCA.HFOUTEN` pin thus allowing to disable it. This makes no sense from practical point of view as it is irreversable but allows to establish minimal FPGA core current consumption.

With the clock turned off completely the power consumption on the `FPGA_VCC1/2/3` rail dropped to ~7mA. This is the bottomline.

tcal-x:
@mkurc-ant , @danc86 , @davidlattimore , this has been subsumed by other work and is obsolete, correct?   Shall I close it?

danc86:
Indeed, PR #597 was the final version of this which went in. We can close this one. Thanks for your earlier work on it @mkurc-ant :-)

PR #529, opened 2022-04-18
conf: ignore files updated by Symbiflow install

alanvgreen:
conf/requirements.txt and conf/environment.yml are updated by `make
sf-install`. `git status` then records them as modified, but the
developer mostly likely does not want to send these updates upstream.

This change adds these two files to .gitignore.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -101,3 +101,5 @@ third_party/yosys
 
 # Environment
 env
+conf/environment.yml
+conf/requirements.txt


tcal-x:
Thanks @alanvgreen!   The way I set this up was a quick hack that I meant to remove when I had a chance.  This is definitely an improvement over the current state and improves things until I can do a proper fix.   

There's some issue with the SiFive certificate (for downloading the riscv toolchain).  Hopefully it will be resolved soon.

Issue #539, opened 2022-04-20
scripts/install_oxide failure: yosys plugins not built


alanvgreen:
I made a fresh checkout of CFU-Playground, then ran `scripts/setup` and `scripts/install_oxide`. It failed with an error message: `../Makefile_plugin.common:47: *** "Didn't find 'yosys-config' under '/usr'".  Stop.`

Running install_oxide a second time gives the same error:
```
~/src/CFU-Playground (main %|u=)
14:07:04 $ scripts/install_oxide
Sudo Installing pre-requisites
[sudo] password for avg: 
Hit:1 https://rapture-prod.corp.google.com goobuntu-workstation@avg920.syd.corp.google.com InRelease
Hit:2 https://rapture-prod.corp.google.com glitter@avg920.syd.corp.google.com InRelease
Get:3 https://rapture-prod.corp.google.com goobuntu-puppet-addons InRelease [8,137 B]
Hit:4 https://rapture-prod.corp.google.com glinux-rodete-base-20220323.03.06 InRelease
Hit:5 https://rapture-prod.corp.google.com glinux-rodete-base-priority-20220323.03.06 InRelease
Hit:6 https://rapture-prod.corp.google.com glinux-rodete-canaries-20220323.03.06@avg920.syd.corp.google.com InRelease
Hit:7 https://rapture-prod.corp.google.com glinux-rodete-extra-20220323.03.06 InRelease
Hit:8 https://rapture-prod.corp.google.com glinux-rodete-extra-priority-20220323.03.06 InRelease
Hit:9 https://rapture-prod.corp.google.com typescript-stable InRelease
Hit:10 https://rapture-prod.corp.google.com usblock-stable InRelease
Get:11 https://rapture-prod.corp.google.com goobuntu-puppet-addons/main amd64 Packages [20.7 kB]
Get:12 https://rapture-prod.corp.google.com goobuntu-puppet-addons/main i386 Packages [19.6 kB]
Fetched 48.4 kB in 3s (13.9 kB/s)                           
Reading package lists... Done
Building dependency tree... Done
Reading state information... Done
1 package can be upgraded. Run 'apt list --upgradable' to see it.
Reading package lists... Done
Building dependency tree... Done
Reading state information... Done
bison is already the newest version (2:3.8.2+dfsg-1).
clang is already the newest version (1:13.0-54).
cmake is already the newest version (3.22.1-1+build1).
flex is already the newest version (2.6.4-8).
[...]

PR #540, opened 2022-04-20
scripts/install_oxide: fixes

alanvgreen:
Contains two fixes:

1. Passes full path to Yosys src the plugin make command, which enables
   it to find the yosys-config script.
2. Passes DESTDIR to the first yosys make invocation so that it can be
   used when constructing the yosys-config script.

Note that the install_oxide script will require a change to the yosys
makefile in order to construct the correct yosys-config script.


alanvgreen:
This is a partial fix for #539

tcal-x (approved):
LGTM

PR #544, opened 2022-04-21
scripts/install_oxide: set installation paths

alanvgreen:
Updates build commands for Yosys, Yosys plugins and Nextpnr to inform
the builds of the final installation directory, under
third_party/usr/local.

In particular,

- for Yosys, replaces use of "DESTDIR=foo", with "PREFIX=foo".
- for Nextpnr, set CMAKE_INSTALL_PREFIX to specify from where the built
  binaries will run.

### DIFF
--- a/scripts/install_oxide
+++ b/scripts/install_oxide
@@ -67,39 +67,39 @@ else
 
 fi
 
-DESTDIR=${CFU_ROOT}/third_party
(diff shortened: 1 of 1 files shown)


alanvgreen:
#539

tcal-x (approved):

PR #548, opened 2022-04-22
hps_model: add new models

alanvgreen:
New presence and second person models from 20220415 and 220220416.


tcal-x (approved):
LGTM

PR #550, opened 2022-04-22
 Add minimal+cfu Vex variant; has the mcycle csr.

tcal-x:
I had to modify some CFU Playground test code that wouldn't compile due to explicit div/divu/rem/remu assembly instructions.  The minimal variant is rv32i and so doesn't expect any type of mul/div/rem instruction.


ShvetankPrakash (approved):
Small change, looks good to me!

PR #553, opened 2022-04-25
Make 'fomu' cpu variant available everywhere incl. sim.

### LINKED ISSUE #543: Specifying CPU variant in verilator
Specifying the cpu variant in a Verilator simulation is not working:
`make PLATFORM=sim load EXTRA_LITEX_ARGS="--cpu-variant=fomu"`
Error says --cpu-variant is not a valid parameter.

tcal-x:
Before, the patching for the fomu cpu variant was only done
when you were targeting the fomu board.  This change
makes it so the patching is always done, so that the fomu
variant is always available, including in Verilator sim.

This fixes #543.

### DIFF
--- a/soc/board_specific_workflows/ice40up5k.py
+++ b/soc/board_specific_workflows/ice40up5k.py
@@ -23,30 +23,6 @@ from litex.soc.integration import builder
 from typing import Callable
 
 
-def patch_cpu_fomu_variant():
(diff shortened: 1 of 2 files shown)


tcal-x:
@alanvgreen @ShvetankPrakash  PTAL

danc86 (approved):

PR #584, opened 2022-05-13
Update scanprof.py to parse prof (per-layer cycles) correctly.

tcal-x:
This version had been on the demo branch.


ShvetankPrakash (approved):
Small changes that look good but will look at in Colab as well!

PR #587, opened 2022-05-14
Fix install_oxide script w.r.t. dsp-ff plugin

tcal-x:
The dsp-ff plugin Makefile now has building a directory
as the default action, which isn't enough, so we must
now explicitly "make all" instead of just "make".

It is currently broken either in CI or if you're running 
`./scripts/install_oxide` locally.


alanvgreen (approved):
lgtm

PR #590, opened 2022-05-20
Ice40 placement

tcal-x:
Generate files `critpath.png` and `critpath.log` in the build `gateware/` directory.   The PNG file contains a color-coded plot of placement color coded by LiteX, CPU, and CFU; the plot is overlayed with the critical path.   The LOG file contains the text list of cells on the critical path, and also the color key for the placement.

To do: generalize to make it easy to adapt for ECP5, Nexus, or any other part that uses Nextpnr.

### DIFF
--- a/scripts/critpath.py
+++ b/scripts/critpath.py
@@ -1,0 +1,119 @@
+#!/bin/bash
+# Copyright 2022 The CFU-Playground Authors
+#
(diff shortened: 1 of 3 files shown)


tcal-x:
We might need to do an `apt install` of matplotlib if not using Conda.

PR #592, opened 2022-05-23
Rebuild local Vex verilogs with #582 fix.

tcal-x:


tcal-x:
@davidlattimore reported that hps_accel on proto2 did not function with this PR, and narrowed down the apparent introduction of an incompatibility to this commit: https://github.com/SpinalHDL/VexRiscv/commit/b2e61caf9e62d8febe59eaf654856edec73e5412.

I tried `mnv2_first` on Arty with this LiteX/VexRiscv/CfuPlugin bump, and it also seems to be malfunctioning.

--> I will see if the issue also shows up in Verilator simulation, and if so, add a regression test for it.

tcal-x:
In fact the `proj/mnv2_first` project malfunction shows up even on current main, since that build directly uses an upstream VexRiscv build.

I've also confirmed that the same malfunctions also occur in Verilator.

Specifically, the malfunctions are:
* project menu --> 1x1 test --> hang
* TFLM unit tests --> hang 
* TFLM models --> mnv2 model --> any --> hang after 3 dots

tcal-x:
Closing without merging; this is obsolete (superseded by #639).

PR #597, opened 2022-05-31
Dynamic clock control between CPU and CFU

robertszczepanski:
After tests done in #514 we want to use a possibility of disabling/enabling clocks to minimize power consumption during inference.

This PR introduces dynamic clock control between CPU and CFU. When CPU receives a confirmation that a command is accepted (`cmd.ready` asserted) its clock gets disabled until CFU is ready to response (`rsp.valid` asserted). Then CPU's clock is enabled and CFU's clock is disabled.

Right now this is a draft since even though clock control between SoC and CFU works, for some reason CFU hangs after receiving a command from CPU. Maybe it requires few additional cycles for setup at system boot?

There is also a weird anomaly - after boot, power consumption is about 44-45mW (~38mW with CFU clock disabled) but after running some tests (e.g. `1: TfLM Models menu -> 1: HPS models -> g: Golden tests (check for expected outputs)`) power consumption rises to around 54mW and stays on that level after each operation (it goes down to 42-44mW during other executions of CFU functions e.g. HPS golden tests, but it rises again after finish). It falls back to 44mW after asserting reset signal in CFU clock domain.


robertszczepanski:
I just pushed changes introducing a proper control of CFU and CPU clocks. Power consumption is reduced to 39-40mW when there is no CFU instruction executed and clocks are switching with each other during CFU runtime.

There will be more tests done to measure properly power consumption during clock switching but in general it's the very first version that controls clocks dynamically.

robertszczepanski:
I simplified code a bit to make it more generic. Right now it should be possible to build all samples using code in this PR.

I also noticed that CI fails and I tested that it's caused by outdated `nextpnr-nexus` version. Locally I'm working on `0.3-31`, CI works on `0.3-21`. Updating it is essential since there were changes needed for DCCs to work. It might also be required to update `prjoxide`. There 2 PR applied during development of this feature:
- nextpnr - https://github.com/YosysHQ/nextpnr/pull/992
- prjoxide - https://github.com/gatecat/prjoxide/pull/27

robertszczepanski:
Both `nextpnr-nexus` and `prjoxide` should be up to date in `LiteX-Hub` right now, so if you could restart CI it will most likely be green.

FYI @tcal-x @alanvgreen

robertszczepanski:
Earlier I assumed that tests on main branch are broken and that these won't pass on board, that's why I assumed that dynamic clock control works when system doesn't hang. Fails were caused by incorrect build command (which was wrong in CI hence my mistake, see #613). After noticing this, I run tests with correct command, tests were still failing so it's been fixed now. Currently this branch should pass any tests which was checked on a HPS board.

This PR also introduces a clock control for LRAM. RAM and port A of Arena are enabled with CPU clock, port B of Arena is enabled with CFU clock. These improvements result in following power consumption values. These were measured with use of INA180A1 differential amplifier to achieve high precision:
| **Clock control mode** | **System status** | **Power consumption** | **Duty cycle (CPU)** | **Duty cycle (CFU)** |
|------------------------|-------------------|-----------------------|----------------------|----------------------|
| Always enabled         | Idle              | 51 mW                 | 100%                 | 100%                 |
| Always enabled         | Acceleration      | 46 mW                 | 100%                 | 100%                 |
| Dynamic switch         | Idle              | **28 mW**             | 100%                 | 0%                   |
| Dynamic switch         | Acceleration      | **30 mW**             | ~43%                 | ~64%                 |

**Idle** means that values were measured when system was initialized with responsive menu but without executing any tests.
**Acceleration** means that values were measured during tests from `1: TfLM Models menu -> 1: HPS models -> g: Golden tests` for default model. It was also checked if tests pass for different reinitialized models and for test layers in `3: Project menu`.

soc/hps_soc.py
```diff
@@ -54,6 +55,72 @@
...
+        self.submodules.fsm = fsm = FSM()
+        # It's just an initial state, once left it's not entered again
+        fsm.act("RESET",
+            # Enable both CFU and CPU
+            self.cfu_cen.eq(1),
+            self.cpu_cen.eq(1),
+
+            NextValue(delay, delay + 1),
+
+            # After 10 cycles disable CFU and go to next state
```
danc86, line 87: Where does the 10 cycles come from? I guess this is because the CFU logic does a little bit of work after reset, to get itself ready?

danc86 (approved):
Thanks for all your work on this, Robert! It looks great :star_struck: and the power reduction numbers are very impressive as well!

I'm going to port this over to hps-firmware and see how it goes there.

danc86:
I wanted to see this in action but I can't easily build a bitstream from CFU-Playground right now, because conda is giving me the latest yosys which suffers from https://github.com/YosysHQ/yosys/issues/3416.

I'm trying to port the changes over to hps-firmware so I can try them there, and I'm seeing something strange.

When I apply the complete set of changes from this PR, it produces a non-working bitstream. I picked apart the changes, starting from introducing the new `osc` clock domain, and with some trial and error I was able to narrow it down. The new clock domain, and the DCC to feed it, are fine on their own. But when I add in this change, to use `osc` for power-on reset:

```
         # Power On Reset
         por_cycles = 4096
         por_counter = Signal(log2_int(por_cycles), reset=por_cycles - 1)
-        self.comb += self.cd_por.clk.eq(self.cd_sys.clk)
+        self.comb += self.cd_por.clk.eq(self.cd_osc.clk)
         self.sync.por += If(por_counter != 0, por_counter.eq(por_counter - 1))
         self.specials += AsyncResetSynchronizer(
-            self.cd_sys, (por_counter != 0))
+            self.cd_osc, (por_counter != 0))
```

then the FPGA never comes out of reset.

In fact if I leave the POR clock fed by `sys` and not `osc` then the rest of the changes here seem to work.

It seems to me that it should be equivalent, whether the POR logic is fed by `osc` or `sys` clocks. During the first 4096 cycles both clocks should be enabled. And `osc` is always enabled, so it *should* be the one to feed the POR counter. I don't understand why it would work with `sys` but not `osc`, nor why you didn't run into the same issue. Could I be using an out-of-date version of migen that is missing some changes in AsyncResetSynchronizer?

robertszczepanski:
I don't think it's related to `AsyncResetSynchronizer` since it's imported form migen actually and I see that it's been last updated a long time ago: https://github.com/m-labs/migen/blob/master/migen/genlib/resetsync.py.

I just created a new environment here with:
```
cd CFU-Playground
make clean env
```
and I rebuilt software and bitstream with:
```
cd proj/hps_accel
PLATFORM=hps EXTRA_NEXTPNR_ARGS="--placer-heap-timingweight 25" EXTRA_LITEX_ARGS="--extra-nextpnr-params" 
make clean software bitstream
```
This works for me.
I'm using additional nextpnr parameter `--placer-heap-timingweight 25` because nextpnr tends to get stuck on routing step (iterating without any progress). If that happens you might want to change this parameter to some other value. For me it always worked for one of `25`, `52` or `79`. Note that parameters must be provided before calling `make` in executed command to avoid overwriting built-in project parameter setup.

Just to be one the same page, here are versions I'm using:
```
$ prjoxide --version
prjoxide 0.1
$ conda list | grep prjoxide
prjoxide                  0.0_430_gd1fc5cd 20220708_203356    litex-hub
$ nextpnr-nexus --version
nextpnr-nexus -- Next Generation Place and Route (Version 0.3-54-g664cec54)
$ yosys --version
Yosys 0.19+18 (git sha1 d25f349f4 [...])
```

That would also mean I'm using one of latest Yosys and I don't have any issues with running a bitstream on a board.

tmichalak:
@danc86 did the bitstream from CFU Playground work for you or you just went straight to testing it with hps-firmware?

soc/hps_soc.py
robertszczepanski: Actually I've never checked that precisely until now. I just assumed that it requires some time to set up. After running some tests it looks like it doesn't need any setup time for most functions. In fact the only thing that fails without these additional cycles is a CFU ping test :) These results look like that because I did test reset executed by software which takes too many cycles from reset to next instruction to precisely check if CFU requires any setup time.

However I noticed that to make setup time relevant for CFU it would require to make `cmd.ready` synchronous which is not the case right now for some reason. Even if CFU is in reset state, CPU won't see it as inavailable if last CFU's state was `WAIT_CMD` since `cmd.ready` will be asserted whole time. Before that, I think there is no point in setting any setup time for CFU. 

I don't think it's something that would bother us right now. Issues don't come up for any tests in current state of CFU Playground and I had to force CFU reset right before calling `cfu_set()` to make ping test fail. If you have any case in hps-firmware that can be affected in some specific case by this issue, then it should be fixed I guess. Otherwise it might not be worth to do it.

danc86:
Actually I went straight to trying these changes in hps-firmware last night, because I was so excited to see them in action :-) but I see from Robert's comment edit that we can explain the mystery of why nobody hit the yosys memory_libmap bug before now.

Today I'll get CFU-Playground running on my HPS proto2 board, using a patched yosys, to confirm Robert's results. And I'll look more closely at the changes I brought into hps-firmware, to see if there might be some mistake I made which could explain the weird behaviour of the POR logic.

soc/hps_lattice_nx.py
```diff
@@ -77,7 +77,7 @@ def initval_parameters(contents, width):
 
 
 class NXLRAM(Module):
-    def __init__(self, width=32, size=128*kB, dual_port=False, init=[]):
+    def __init__(self, width=32, size=128*kB, dual_port=False, init=[], cen={"a": 0b1, "b": 0b1}):
```
danc86, line 80: I was thinking more about this. I think it would be cleaner and more consistent to make the A and B clock enable signals be attributes on this module, rather than passed in at construction. We already have `b_addrs` and `b_douts` attributes, we could add `a_clkens` and `b_clkens` the same way. It would also mean a bit less passing down this cen parameter through various functions.

What do you think?

robertszczepanski: Sure, I'm not a big fan of my approach, I don't consider it intuitive and that passing parameters doesn't look good but I didn't think about that too much. Your idea seems to be better. I'll apply it, thanks.

robertszczepanski:
Looks like CI is failing due to some Conda infrastructure issues, it might be good to restart it in few hours to try again.

I applied a change to LRAM clock enables so these are attributes now. I did the same thing with clock control module's arguments.

Whole clock control logic is now separated to another file. I gave a shot to use Amaranth to write that module but it might need a bit more work to make it compatible with SoC that's written in Migen. I decided to leave it in Migen without digging too much.

There is also a parameter `--dynamic-clock-control` that's added to CI for oxide tests.
Parameter `--cfu-mport` is passed to HpsSoC class right now which should be more intuitive since function dependent on that parameter (`connect_cfu_to_lram()`) was already existing inside of that class anyway.

Whole clock control logic is a bit dependent on these 2 parameters so it should be more generic.

danc86:
Hmm yeah looks like it's getting a 403 from Cloudflare while fetching the conda packages, which is more likely a rate limit than an outage. We should really be caching these packages instead of fetching them all over and over again for every CI build. Anyway, I'll click the button again tomorrow and see if we have more luck.

Apart from that, this PR is looking good to me. Using patched yosys, I see that this is working nicely on HPS proto2, so we can land this now. I'll continue to investigate the weird problem I was seeing with the POR logic.

danc86 (approved):

tcal-x:
Yeah, I'm getting the 403 even when directly downloading the tar file to my laptop:
```
$ wget https://conda.anaconda.org/litex-hub/linux-64/nextpnr-nexus-0.3_65_gdfea9547-20220720_121435_py37.tar.bz2
....
```
Oh it just resolved for me...no more 403, for now anyway.

danc86:
Looks like conda is happy again and the CI has passed, so let's land this. Thanks @robertszczepanski for all your work on it.

danc86:
Just for the record, in the ChromeOS copy of this design, I found that I have to ensure the `sys` clock domain reset is driven from `osc`:

```
self.comb += self.cd_sys.rst.eq(self.cd_osc.rst)
```

otherwise it never comes out of reset properly. The thing I'm unsure about is: why didn't we have the same problem here in CFU-Playground?

danc86:
I suspect we probably *should* be driving the `cfu` clock domain reset signal in the same way, but we can get away without it because the CFU module (which is the only thing in the `cfu` clock domain) gets explicitly fed by the `osc` reset instead.

tmichalak:
@danc86 I think we should make the same change for CFU Playground, even if it is working there now.

danc86:
Indeed, I sent PR #621 to add it.

robertszczepanski:
I've done additional tests for hps-firmware on [main branch](https://chromium.googlesource.com/chromiumos/platform/hps-firmware/+/76d19ef3053ade8143626ac3141d36e83ff1f7c4) and with [dynamic clock control changes](https://chromium-review.googlesource.com/c/chromiumos/platform/hps-firmware/+/3783309). Measurements were taken with INA180A3 differential amplifier and Logic Analyzer, values are calculated from RMS voltage of given timespan. For testing I used `scripts/fpga-rom-run --no-mcp-i2c` script and a `fpga_self_test` command. Acceleration measurement is calculated based on time from the beginning to the end of `fpga_self_test`.

| **Clock control mode** | **System status** | **Power consumption** |
|------------------------|-------------------|-----------------------|
| Always enabled         | Idle              | 88 mW                 |
| Always enabled         | Acceleration      | 77 mW                 |
| Dynamic switch         | Idle              | **49 mW**             |
| Dynamic switch         | Acceleration      | **55 mW**             |

FYI @danc86 @tcal-x @tmichalak

PR #600, opened 2022-06-13
Delete index.html

ShvetankPrakash:
Do not need this here. It was initially added to test if we could host our workshop site. Instead, created `gh-pages` branch created for hosting workshop site and added our site there.


tcal-x (approved):
LGTM!
Can be merged when tests finish and pass.

PR #601, opened 2022-06-13
hps_accel: Latch output of adding input_offset

davidlattimore:
This shortens a long path and allows model eval to function correctly
when div=5.


alanvgreen (approved):
Very nice.

PR #608, opened 2022-07-05
hps_accel: Support shifting right by 2

davidlattimore:
This is needed for a new model that we'll be using.


danc86 (approved):
LGTM but it would be nice if we could also have a matching assertion on the C++ side, for the shift values we can handle here, to prevent this problem in future.

davidlattimore:
Turns out we need to support shifting right by 12 as well, so I updated the commit to include that.

I also added a commit that checks if the shift is within the supported range.

danc86 (approved):
Very nice

PR #609, opened 2022-07-12
RoundingDivideByPowerOfTwo: Simplify threshold bit pattern

davidlattimore:
Instead of patterns like 11, 111, 1111 for positive values and 100,
1000, 10000 for negative values, we use 100, 1000, 10000 for positive
values and 101, 1001, 10001 for negative values. The new bit patterns
are simpler to compute.

We then compensate for the increased threshold by using >= instead of >.

In hps-firmware, this reduces LUT usage from 8420 to 8373 (-47) and
increases fmax from 97.28 to 101.56.


danc86 (approved):
Wow

PR #613, opened 2022-07-15
HPS build command fix/improvement

robertszczepanski:
I think there are 2 unintended issues in HPS build command:

1. Each time you call `make PLATFORM=hps bitstream`, a bitstream is actually build twice, first time with `--no-compile-software` flag and second time without it. It looks like a leftover after some conflicts on merge/rebase. Removing that makes building bitstream twice faster.

2. If you use `--separate-arena` argument, you have to provide `--cfu-mport` in order to connect that arena to CFU and pass any tests. I've added this to all CI builds.

I was also wondering if you still want to build CIs with a `--cpu-variant` argument since you recently added `hps+cfu` as a default for HPS but I didn't introduce this change into PR. If you want me to do it, I can add another commit about that.

FYI @tcal-x


danc86 (approved):
Thanks for these fixes Robert

PR #615, opened 2022-07-20
models/hps: add latest "shared" model

danc86:
Taken from:
https://chromium.googlesource.com/chromiumos/platform/hps-firmware/+/ec8183f8e3c117a91e23dccbd463ba0e848ae70d

Unlike the earlier HPS models, this model produces two scores, which
necessitates some unfortunate contortions in the golden test code.

Also load the latest model by default, since it's the most important one
to run golden tests against.


danc86:
An alternative to this ugly mess in the golden test code, would be to just delete the older models and always expect two scores.

davidlattimore (approved):

PR #619, opened 2022-07-24
Bump dependencies including Lite* and VexRiscv.

tcal-x:
Local VexRiscv verilogs HAVE NOT been rebuilt;
they will be updated in a subsequent PR.


alanvgreen (approved):

danc86 (approved):
Thanks for doing this tcal, it will be a very useful reference for us when we go to do the same uprev in ChromeOS

tcal-x:
Tested with HPS on board -- works.  

The only issue that I found is that the KWS project no longer fits on Fomu (roughly 80 LCs over capacity).

PR #621, opened 2022-07-26
hps: drive sys reset signal from osc reset signal

danc86:
When I was porting the dynamic clock control change (PR #597) to ChromeOS:
https://chromium-review.googlesource.com/c/chromiumos/platform/hps-firmware/+/3783309
I found that I needed to connect the sys clock domain's reset signal to the osc clock domain's reset signal. Otherwise it doesn't seem to come out of reset properly.

I'm not sure why we got away without this in CFU-Playground, but let's add it here for consistency too.

### DIFF
--- a/soc/hps_proto2_platform.py
+++ b/soc/hps_proto2_platform.py
@@ -79,6 +79,7 @@ class _CRG(Module):
             o_CLKO=ClockSignal("sys"),
             i_CE=self.sys_clk_enable,
         )
+        self.comb += self.cd_sys.rst.eq(self.cd_osc.rst)
         # We make the period constraint 7% tighter than our actual system
         # clock frequency, because the CrossLink-NX internal oscillator runs
         # at ±7% of nominal frequency.


robertszczepanski:
LGTM, tested this on board with `proj/hps_accel` with and without dynamic clock control.

There's just one thing about this PR, it causes a known ABC issue to occur if I build it without `--dynamic-clock-control`: https://github.com/berkeley-abc/abc/issues/165. Personally I apply some permutation to the design and it's gone, in this case it's enough to add:
```
sig = Signal()
self.comb += sig.eq(0)
```
I'm not sure though if you want to apply such workaround on main branch, this might become useless or even it might become a bug trigger itself after any new design modification.

tcal-x:
> ...it causes a known ABC issue to occur ... [berkeley-abc/abc#165](https://github.com/berkeley-abc/abc/issues/165). 

Yes I've been seeing this recently as well (even without the changes in this PR).   My workaround has been to set the clock div to 10 (instead of 9).   As you say, we just need to perturb the design a little to avoid the issue.

soc/hps_proto2_platform.py
```diff
@@ -79,6 +79,7 @@ def __init__(self, platform, sys_clk_freq):
             o_CLKO=ClockSignal("sys"),
             i_CE=self.sys_clk_enable,
         )
```
tcal-x, line 81: For my understanding, is the clock of `cd_sys` driven by the clock of `cd_osc` via this instantiation?

robertszczepanski: Indeed, DCC guarantees no glitches and equality in phase of its input clock signal and output signal. According to [Migen documentation](https://m-labs.hk/migen/manual/reference.html#migen.fhdl.structure.ClockSignal), each clock signal retrieved by `ClockSignal()` refers to the same signal so `sys` domain should be identical to `osc` domain but it also has an enable feature.

tcal-x:
I looked at the generated Verilog *without* this change, and `sys_rst` is tied to 1'b0.   There's a large `if` statement zeroing out LiteX registers that will never execute:
```
       if (sys_rst) begin
                reset_storage <= 2'd0;
                reset_re <= 1'd0;
                scratch_storage <= 32'd305419896;
                scratch_re <= 1'd0;
                bus_errors_re <= 1'd0;
                bus_errors <= 32'd0;
                litespiddrphycore_en <= 1'd0;
                litespiddrphycore_count <= 4'd11;
                etc.
```
`sys_rst` also drives the `reset` inputs of the CPU and CFU!

So yes, it is somewhat surprising that it works at all without this change.

tcal-x (approved):

danc86:
The DCO bot reminded me I forgot to put a Signed-off-by footer on my commit. Amended to add the footer.

PR #623, opened 2022-07-27
Disable local install of Yosys binary.

tcal-x:
Use Conda if you need Yosys.


danc86 (approved):

tcal-x:
The large number of "Test projects" Renode tests were relying on Yosys for their Amaranth processing; they don't currently use Conda.

Before we switch these tests over to using Conda, I will probably look into this lead that @danc86 found: https://github.com/conda-incubator/setup-miniconda#caching-environments to use caching to improve Conda performance; it might also help with some likely rate limiting we saw recently.

Also, I will probably create another environment configuration that just has Yosys and whatever else is needed for these Renode CI tests.   

Then potentially if we can assume that we're always using Conda, we might be able to simplify some of the other dependencies that we install either with `apt` or by cloning (e.g. `meson`).

mithro:
FYI - @PiotrZierhoffer

tcal-x:
FYI @PiotrZierhoffer @robertszczepanski -- I updated the test-projects workflow to use Conda (with caching).   The individual tests no longer get the riscv32 gcc toolchain using wget from sifive.   They now get it via Conda.

danc86 (commented):
Nice!

PR #624, opened 2022-07-27
Add dynamic clock control for CFU Arena ports

robertszczepanski:
This PR introduces dynamic clock control for CFU LRAM Arena ports. It's based on fetcher's start and stop signals which should correspond to LRAM read and idle states.

After looking at waveforms it looks like parts where CFU is enabled, for significant part of time, match parts between memory access start and stop. As a result additional logic disables LRAM clock for a very little additional time in comparison to CFU clock enable itself.

About power consumption, there is no significant difference between this PR and main branch. Differences vary about +/- 1mW depending on quality of placement/routing done by nextpnr (different seeds give different results) so it's hard to tell if it's measurement error, bad nextpnr seed or if it's real improvement.


robertszczepanski:
I did play a bit with `first`/`last` signals in accelerator fetchers to maximize time when LRAM is disabled but I couldn't get this to work. All changes made won't benefit in any significant improvements. Looking at waveforms I can see that there are some additional cycles when LRAM is disabled but it breaks tests and it's not an amount easy to notice. I think in current state of CFU this might be the best we can achieve about LRAM clock control.

danc86:
Thanks very much for trying this out, Robert. It was certainly a worthwhile experiment. But I'm not surprised that the accelerator needs access to the LRAMs for the majority of its operational time. It seems like the conclusion here is that the negligible benefit is not worth the extra complexity this would add, is that right?

robertszczepanski:
Yes, I think it's not worth to add complexity when it looks like there's no real benefit from that (or it's so small that it's hard to tell if it's the patch or if it's just different placement/routing).

PR #627, opened 2022-07-28
Revert "Fix compilation if no TFLM models included in Makefile"

tcal-x:
This change causes problems; I think it makes the model version of `no_menu` more visible so that it clashes with the embench version of `no_menu`.

If you do this:
```
% cd proj/proj_template_v
% make software
```

you will get:
```
  ...
  CXX tflite_unit_tests.cc      tflite_unit_tests.o
  AS  crt0-vexriscv.S   crt0-vexriscv.o
  LD       software.elf
/media/tim/GIT/tcal-x/CFU-Playground/env/conda/envs/cfu-common/bin/../lib/gcc/riscv32-elf/10.1.0/../../../../riscv32-elf/bin/ld: src/third_party/embench_iot_v1/embench.o: in function `no_menu':
/media/tim/GIT/tcal-x/CFU-Playground/proj/proj_template_v/build/src/third_party/embench_iot_v1/embench.c:23: multiple definition of `no_menu'; src/models/models.o:/media/tim/GIT/tcal-x/CFU-Playground/proj/proj_template_v/build/src/models/models.c:32: first defined here
collect2: error: ld returned 1 exit status
make[2]: *** [Makefile:164: software.elf] Error 1
make[2]: Leaving directory '/media/tim/GIT/tcal-x/CFU-Playground/proj/proj_template_v/build'
make[1]: *** [../proj.mk:232: /media/tim/GIT/tcal-x/CFU-Playground/proj/proj_template_v/build/software.bin] Error 2
make[1]: Leaving directory '/media/tim/GIT/tcal-x/CFU-Playground/proj/proj_template_v'

```

### DIFF
--- a/common/src/models/models.h
+++ b/common/src/models/models.h
@@ -23,7 +23,6 @@ extern "C" {
 
 // For integration into menu system
 void models_menu();
-void no_menu();
 
 #ifdef __cplusplus
 }


danc86 (approved):

common/src/models/models.h
```diff
@@ -23,10 +23,9 @@ extern "C" {
 
 // For integration into menu system
 void models_menu();
-void no_menu();
```
ShvetankPrakash, line 26: Instead of reverting and removing this can we just change it to `no_models_menu`. And then in the embench I can change it from `no_menu` to `no_embench_menu`.

ShvetankPrakash: Because removing this change technically handles this issue of the `no_menu` conflict during the build, but leaves the bug still of `no_menu` prototype missing if you include no models in the defines of the project Makefile.

tcal-x:
Hi @ShvetankPrakash , I think the compile time error was the intended behavior, with the motivation, why would you build with TFLM and without any models?   Perhaps there is a use case, but I think that was the original thinking.   If you don't want TFLM, then instead use the SKIP_TFLM=1 mechanism.

If we do want to avoid the compile-time error, to allow compiling with TFLM and without any models, then I think we should use a different approach.   I'm going to revert then we can take another look.

ShvetankPrakash:
> Hi @ShvetankPrakash , I think the compile time error was the intended behavior, with the motivation, why would you build with TFLM and without any models? Perhaps there is a use case, but I think that was the original thinking. If you don't want TFLM, then instead use the SKIP_TFLM=1 mechanism.
> 
> If we do want to avoid the compile-time error, to allow compiling with TFLM and without any models, then I think we should use a different approach. I'm going to revert then we can take another look.

Sounds good to me for now @tcal-x . I see what you're saying. I think I had interpreted the intended behavior differently. Based on the message to be printed out in the model menu in the case that there are no models, I thought you would want it to compile still but print out this [message](https://github.com/google/CFU-Playground/blob/e616742b5a57fb1acbb599f65f37d8a453e21b4e/common/src/models/models.c#L75).

ShvetankPrakash:
> Hi @ShvetankPrakash , how exactly did you arrive at the original compile-time error? I'm not seeing it (after merging this revert).
> 
> I commented out the PDTI8 model in `proj/proj_template_v/Makefile`. `make software` works for me, as well as `make renode` (no compilation error).
> 
> Using `make renode` and going to the model menu, I see the `No models selected!` message.
> 
> nomodel

Hmm, when I was running on the board yesterday and commented out the PDIT8 model I was getting the following error:
```…
LD       software.elf
/home/shvetank/Repos/CFU-Playground/env/conda/envs/cfu-symbiflow/bin/../lib/gcc/riscv32-elf/10.1.0/../../../../riscv32-elf/bin/ld: src/models/models.o: in function `.LANCHOR0':
models.c:(.sdata.MENU+0x10): undefined reference to `no_menu'
collect2: error: ld returned 1 exit status
make[3]: *** [Makefile:164: software.elf] Error 1
make[3]: Leaving directory '/home/shvetank/Repos/CFU-Playground/proj/proj_template/build'
make[2]: *** [../proj.mk:232: /home/shvetank/Repos//CFU-Playground/proj/proj_template/build/software.bin] Error 2
make[2]: Leaving directory '/home/shvetank/Repos/CFU-Playground/proj/proj_template'
```
But I will have to re run now and see since it seems to be working fine for you. I am running some validation tests on the board at the moment that take a while but will try to run this again after.

PR #630, opened 2022-07-28
Add json-c and libevent to conda env.

tcal-x:
I think they used to be implicitly loaded as dependencies of the Verilator package, but now they need to be explicitly specified.

### DIFF
--- a/conf/environment-symbiflow.yml
+++ b/conf/environment-symbiflow.yml
@@ -21,6 +21,8 @@ dependencies:
   - lxml
   - simplejson
   - intervaltree
+  - json-c
+  - libevent
   - python=3.7
   - pip
   - pip:
--- a/conf/environment.yml
+++ b/conf/environment.yml
@@ -14,6 +14,8 @@ dependencies:
   - litex-hub::iceprog
   - litex-hub::yosys
   - litex-hub::symbiflow-yosys-plugins
+  - libevent
+  - json-c
   - python=3.7
   - pip
   - pip:


ShvetankPrakash:
LGTM

ShvetankPrakash (approved):

mithro:
@tcal-x - Can you log a bug on the Verilator package that these should be runtime dependencies that are automatically installed when you request Verilator?

tcal-x:
> @tcal-x - Can you log a bug on the Verilator package that these should be runtime dependencies that are automatically installed when you request Verilator?

Sure!

tcal-x:
I tried using https://github.com/tcal-x/verilator-make-env-test as a test case to prove that `json-c` and `libevent` were required for Verilator, but I ended up proving the opposite.   The small Verilator example there can compile *and link* without `libevent` or `json-c`, so I guess they are not strictly necessary for Verilator, but just for LiteX's simulation using Verilator.   So I think it's OK that they're not listed as dependencies for the Verilator package itself.  @mithro

mithro:
@tcal-x - It seems like we might need a conda package like `litex-sim` which depends on `litex`, `json-c`, `libevent` and similar stuff?

tcal-x:
@mithro, I'm finally getting back to this...I've confirmed it's `litex_sim` that has these extra dependencies --- they're required even running  a pure LiteX sim build.   In fact this is on the LiteX home page https://github.com/enjoy-digital/litex:
```
$ sudo apt install libevent-dev libjson-c-dev verilator
$ litex_sim --cpu-type=vexriscv
```
I'll open an issue on hdl/conda-eda.

Issue #596, opened 2022-05-25
Latest VexRiscv CFU interaction is hanging


tcal-x:
When bumping VexRiscv and pre-generated verilog to recent master, we saw hangs using the CFU using either locally generated verilog files, or those generated upstream.    Original discussion here: https://github.com/google/CFU-Playground/pull/592#issuecomment-1136534363.   @davidlattimore had found the commit that seemed to cause the issue: https://github.com/SpinalHDL/VexRiscv/commit/b2e61caf9e62d8febe59eaf654856edec73e5412

Further investigation shows that it seems to be due to a new "en" register, initialized `false`, accessed as bit 31 of a csr, used here:
https://github.com/SpinalHDL/VexRiscv/commit/b2e61caf9e62d8febe59eaf654856edec73e5412#diff-0c3bf8c460b84c679bb5ae6755da7da8487141f0ffb5cadcc76d8775e024d169R184

If I start with a new verilog that exhibits the hanging, and then hand-modify all initialization and modification of verilog register `CfuPlugin_en` so that it is only set to 1'b1, then the hanging behavior goes away.

I suspect that we just need to write `true` to this CSR bit during software initialization.

However, this exposed a gap in our CI testing -- we have no tests of the VexRiscv / CfuPlugin RTL.   We need to add one or more tests before closing this issue.

tcal-x:
From the spec at https://github.com/grayresearch/CFU/blob/main/spec/spec.pdf:

> 2.2.1. mcfu_selector CSR 0xBC0: select active CFU and state context
>
> .en: enable custom interface multiplexing
• When en=0, disable custom interface multiplexing. The cfu_id and state_id fields are ignored. No CFU is
selected. Custom-0, custom-1, or custom-2 instructions execute the CPU’s built-in custom instructions.
• When en=1, enable custom interface multiplexing. The cfu_id and state_id fields select the current CFU and
state context. Custom-0/-1/-2 instructions issue CFU requests to the CFU identified by cfu_id and to the state
context identified by state_id.

`.en` is bit 31 of the CSR.

tcal-x:
There seems to be a conflict at CSR 0xBC0:
https://github.com/enjoy-digital/litex/blob/master/litex/soc/cores/cpu/vexriscv/csr-defs.h#L6
https://github.com/SpinalHDL/VexRiscv/blob/master/src/main/scala/vexriscv/plugin/CfuPlugin.scala#L103

[...]

PR #639, opened 2022-08-02
Update VexRiscv: no rw check on blockmem; work around #596 (CFU CSR collision)

tcal-x:
VexRiscv build script updated and all verilog files rebuilt.  Makefile fixed (added missing verilog targets).

Now at tip of master of the VexRiscv.  The issue of #596 is avoided by configuring the CfuPlugin with "withEnable = false".     See also https://github.com/SpinalHDL/VexRiscv/issues/258.
    
This also incorporates a change to add the attribute   `(* no_rw_check *)`  on the block memories in VexRiscv --     regfile and caches.   This prevents Yosys from adding circuitry for a check that we don't need, according to Charles P.    Having the check consumes resources, particularly noticeable     when targetting ice40.
    

davidlattimore (approved):

danc86 (approved):

tcal-x:
Note: `no_rw_check` did NOT have any effect on resource consumption in HPS builds.

PR #644, opened 2022-08-09
Dse framework

ShvetankPrakash:
PR for design space exploration framework as an addition to CFU-Playground's capabilities. Using this framework we can explore different soft core configurations and look at the trade-offs for a given workload, enabling exploration of CFU vs CPU into the design space as well.


soc/sim.py
```diff
@@ -137,6 +139,7 @@ def main():
         build=args.run,
         run=args.run,
         sim_config=sim_config,
+        interactive=False,
```
tcal-x, line 142: Does this prevent interactive sim (`make PLATFORM=sim load`)?   Do we need to make this configurable?

tcal-x:
Replaced by #647

PR #647, opened 2022-08-12
Redo dse framework

ShvetankPrakash:
PR for design space exploration framework as an addition to CFU-Playground's capabilities. Using this framework we can explore different soft core configurations and look at the trade-offs for a given workload, enabling exploration of CFU vs CPU into the design space as well. Note: this replaces PR #644 by adding signoff.


tcal-x (approved):
LGTM

PR #648, opened 2022-08-12
Unwedge TFLM

tcal-x:
Automatic updates from TFLM had been stuck for a while.   Some APIs changed upstream, and corresponding changes needed to be made in code local to CFU Playground.   This PR makes those necessary changes.

* One of the conv2d kernels template-ized the type of the bias data.   This was complicated by the fact that in the CFU Playground specialization of conv2d, we had moved the guts of the kernels from `conv.h` to `conv.cc`.

* There was a minor type change in `micro_time.cc`, from int32 to uint32.

This PR includes the latest automatic update of TFLM, plus the required local updates.


PiotrZierhoffer (approved):
I'd say your changes are not controversial in any way, lgtm

PR #675, opened 2022-09-12
Add 'donut' demo from LiteX.

tcal-x:
It is enabled in projects `skip_tflm`, `proj_template`, and `proj_template_v`.
It *can* be enabled in any project.

### DIFF
--- a/common/src/main.c
+++ b/common/src/main.c
@@ -36,6 +36,9 @@
 #include "models/models.h"
 #endif
 
+void donut(void);
+
+
 #ifdef PLATFORM_sim
 static void trace_sim() {
   puts("Beginning trace");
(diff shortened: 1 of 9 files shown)


common/src/main.c
```diff
@@ -36,6 +36,9 @@
 #include "models/models.h"
 #endif
 
+void donut(void);
```
ShvetankPrakash, line 39: Can we create a `donut.h` header file so we can `#include "third_party/litex-donut/donut.h" instead of having this one-off prototype hear since none others are listed in this file like this?

tcal-x: Sure; I was perhaps being too faithful to the original source (that's how they do it in the LiteX demo, which has a similar menu-driven structure).

ShvetankPrakash (approved):
LGTM

Issue #715, opened 2022-10-22
In our build, MicroPrintf() doesn't understand "%lu" and prints "lu" instead of the number.


tcal-x:
In the following line from `third_party/tflite-micro/tensorflow/lite/micro/micro_profiler.cc`,
```
    MicroPrintf("%d,%s,%" PRIu32, i, tags_[i], ticks);
```
`PRIu32` gets replaced by `lu`, so the format string becomes `"%d,%s,%lu"`, but in the display, the `%lu` is not replaced by the value of `ticks`.  It just prints out "`lu`".   If instead I directly edit the format string to use `%u`, then the value of `ticks` gets printed.

tcal-x:
I filed https://github.com/tensorflow/tflite-micro/issues/1494.

PR #719, opened 2022-10-25
Patch tflite profiler until it's fixed upstream (#715).

tcal-x:


tcal-x:
See #715.   A bug has been filed in TFlite-micro.

danc86 (approved):

PR #727, opened 2022-10-28
Fix avg pdti8

alanvgreen:
Fixes for problems with avg_pdti8 caused by TfLM changes.

Tested with Symbiflow on Arty.


tcal-x (approved):
LGTM!

We should have a test for each project in CI so that we would have caught this sooner.

tcal-x:
@alanvgreen , hmm, the DCO check got crossed up on your signoff e-mail address?

https://github.com/google/CFU-Playground/pull/727/checks?check_run_id=9163822812

alanvgreen:
Ugh. Some time since I last used this confix, someone must have added a check that that the github account's primary email address is being used to sign off. I guess I'll do that from now on.

PR #748, opened 2022-12-21
My tflm fixes

tcal-x:
My manual fixes *on top* of the changes in #734.

Some files/directories moved, so they need to be deleted from their old locations.


tcal-x:
@proppy , this is an example of an automatic TFLM update requiring some manual fixes.   Currently I need to apply a fix for https://github.com/tensorflow/tflite-micro/issues/1494 each time.   The other changes I made here are due to recent changes in the TFLM code.

tcal-x:
I rebased on top of the fix in #749.

tcal-x:
@ShvetankPrakash ping

proj/hps_accel/src/tensorflow/lite/kernels/internal/reference/integer_ops/fully_connected.h
```diff
@@ -93,6 +95,20 @@ inline void FullyConnected(
...
+    const RuntimeShape& bias_shape, const int32_t* bias_data,
+    const RuntimeShape& output_shape, int8_t* output_data) {
+  TFLITE_DCHECK_NE(unpacked_filter_data, nullptr);
+  tflite::tensor_utils::UnpackDenseInt4IntoInt8(
+      filter_data, filter_shape.FlatSize(), unpacked_filter_data);
+  FullyConnected(params, input_shape, input_data, filter_shape,
+                 unpacked_filter_data, bias_shape, bias_data, output_shape,
+                 output_data);
+}
+
```
ShvetankPrakash, line 111: Where was this before/why did we have to move it local code?

proj/hps_accel/src/tensorflow/lite/kernels/internal/reference/integer_ops/fully_connected.h
```diff
@@ -16,8 +16,10 @@ limitations under the License.
 #define TENSORFLOW_LITE_KERNELS_INTERNAL_REFERENCE_INTEGER_OPS_FULLY_CONNECTED_H_
 
 #include <cstdio>
+#include <algorithm>
```
ShvetankPrakash, line 19: @tcal-x is this the line triggering the litex issue btw?

proj/hps_accel/src/tensorflow/lite/kernels/internal/reference/integer_ops/fully_connected.h
tcal-x: This file exists in TFLM.   This is a locally modified version.   What it *should* be is a copy of the current TFLM file, patched with our local changes.   Instead, we maintain a copy of the entire file, which then overlays the TFLM version.   When modifications occur upstream, we need to manually copy them into this copy of the file.

Essentially, I'm manually propagating this change: https://github.com/tensorflow/tflite-micro/commit/1d1d6b968d03c289da6ff9c8384df22aac29391f#diff-2cb25dbe2523a0754af5b5057b800add3642578751a4083dd094b5b99ecdecb2  into our modified copy.

Again, a better way would be to store just the patches for our local changes, so that they could be applied to the latest TFLM version.   That way, most of the time we could automatically combine upstream changes with our local changes (except when there is a conflict).   But that would be a bit more complicated to set up.

proj/hps_accel/src/tensorflow/lite/kernels/internal/reference/integer_ops/fully_connected.h
tcal-x: TFLM uses `<algorithm>` extensively (https://github.com/google/CFU-Playground/search?q=algorithm), so it is not just this one use.   This is just a case where its addition needed to be propagated into one of our local versions.

proj/hps_accel/src/tensorflow/lite/kernels/internal/reference/integer_ops/fully_connected.h
ShvetankPrakash: Thanks for explaining! Got it now.

ShvetankPrakash (approved):
LGTM

PR #757, opened 2023-01-09
Run Serv in CFU Playground

ShvetankPrakash:
This PR adds basic, necessary changes to build & run the baseline Serv implementation (i.e. no MDU, no CFU) in CFU Playground. This is the first addition/use of another CPU in the Playground :)

Here is a test to give Serv a spin. Add the following to the end of the Makefile in `proj/proj_template`:
```
DEFINES += INCLUDE_EMBENCH_PRIMECOUNT
export SERV=1
DEFINES += USE_LITEX_TIMER
```

Then run:
```
make prog EXTRA_LITEX_ARGS="--cpu-type=serv --cpu-variant=standard"
make load EXTRA_LITEX_ARGS="--cpu-type=serv --cpu-variant=standard"
```

Try to run the primecount workload from the Embench benchmark suite and see the following output running on Arty:
```
CFU Playground
==============
 2: Functional CFU Tests
 3: Project menu
 4: Performance Counter Tests
 6: Benchmarks
 7: Util Tests
 8: Embench IoT
main> 8

Running Embench IoT

Embench IoT
===========
 1: primecount
 x: eXit to previous menu
benchmarks> 1

Running primecount
OK  Benchmark result verified
Spent 337298843 cycles
---
```

You can also run TFLM Unit/Golden tests and they should all pass. Note that because Serv does not implement `mcycle` the built in Litex Timer must be used by the user for profiling cycle counts. Functionality to call this timer has been added in the PR, and its use is demonstrated in the Embench workloads.


ShvetankPrakash:
Replaced by #760

PR #760, opened 2023-01-10
Updates to run Serv in CFU Playground

ShvetankPrakash:
This PR adds basic, necessary changes to build & run the baseline Serv implementation (i.e. no MDU, no CFU) in CFU Playground. This is the first addition/use of another CPU in the Playground :)

Here is a test to give Serv a spin. Add the following to the end of the Makefile in `proj/proj_template`:
```
DEFINES += INCLUDE_EMBENCH_PRIMECOUNT
export SERV=1
DEFINES += USE_LITEX_TIMER
```

Then run:
```
make prog EXTRA_LITEX_ARGS="--cpu-type=serv --cpu-variant=standard"
make load EXTRA_LITEX_ARGS="--cpu-type=serv --cpu-variant=standard"
```

Try to run the primecount workload from the Embench benchmark suite and see the following output running on Arty:
```
CFU Playground
==============
 2: Functional CFU Tests
 3: Project menu
 4: Performance Counter Tests
 6: Benchmarks
 7: Util Tests
 8: Embench IoT
main> 8

Running Embench IoT

Embench IoT
===========
 1: primecount
 x: eXit to previous menu
benchmarks> 1

Running primecount
OK  Benchmark result verified
Spent 337298843 cycles
---
```

You can also run TFLM Unit/Golden tests and they should all pass. Note that because Serv does not implement `mcycle` the built in Litex Timer must be used by the user for profiling cycle counts. Functionality to call this timer has been added in the PR, and its use is demonstrated in the Embench workloads. 

This replaces #757.


ShvetankPrakash:
@tcal-x most of the tests were failing in #757 because of a Renode issue that was fixed with a commit. I merged that commit and the rest of main into this branch now. I think some tests are failing now because of the env issue you mentioned.

proj/proj.mk
```diff
@@ -81,6 +81,11 @@ ifdef SKIP_TFLM
 export DEFINES    += SKIP_TFLM
 endif
 
+# Need a slower baudrate when communicating with Serv
+ifdef SERV
+export UART_SPEED = 115200
+endif
+
```
tcal-x, line 88: Can we make this be close to the default definition of `UART_SPEED` -- either move the def above down here, or move this up there, so that everything dealing with `UART_SPEED` is together.

ShvetankPrakash:
@tcal-x fixed `UART_SPEED` and all tests passing btw 👍

tcal-x (approved):
LGTM!

PR #783, opened 2023-02-24
Add --no-std switch to verilator

PiotrZierhoffer:
@tcal-x can you please verify if this fixes the problem for you?


tcal-x:
Hi @PiotrZierhoffer , thanks!  --- that did fix the issue I saw running "make renode" manually, but there are many failing tests now?

PiotrZierhoffer:
It stems from the fact that the CI uses older verilator.

Let's [WIP] this - we'll change the package to implicitly provide this switch, so that the user interface doesn't change.

tcal-x:
If I recall correctly, we originally added Verilator to the list of Conda packages used because the version that we got from `sudo apt install verilator` in Google Colab (Ubuntu 18) was too old.

I thought all of the Github Actions now relied on the Conda packages, which should hide an older system version of Verilator?

PiotrZierhoffer:
We made some changes to the verilator conda package, as it was failing the build on GCP. 

We need to update it a little bit further to address this use case, I believe.

tcal-x:
@PiotrZierhoffer sorry if I dropped the ball here -- what is the status of this?

Issue #808, opened 2023-11-23
Symbiflow not working for Xilinx


navan93:
Similar to the issue #722 I am seeing that synthesis is failing 

```
Project status:
    [S] bitstream:  bitstream -> /home/navaneeth/Projects/openmpw/CFU-Playground/soc/build/digilent_arty.proj_template/gateware/digilent_arty.bit
    [N] build_dir:  /home/navaneeth/Projects/openmpw/CFU-Playground/soc/build/digilent_arty.proj_template/gateware
    [S] eblif:  synth -> /home/navaneeth/Projects/openmpw/CFU-Playground/soc/build/digilent_arty.proj_template/gateware/digilent_arty.eblif
    [S] fasm:  fasm -> /home/navaneeth/Projects/openmpw/CFU-Playground/soc/build/digilent_arty.proj_template/gateware/digilent_arty.fasm
    [S] fasm_extra:  synth -> /home/navaneeth/Projects/openmpw/CFU-Playground/soc/build/digilent_arty.proj_template/gateware/digilent_arty_fasm_extra.fasm
    [S] io_place:  ioplace -> /home/navaneeth/Projects/openmpw/CFU-Playground/soc/build/digilent_arty.proj_template/gateware/digilent_arty.ioplace
    [S] net:  pack -> /home/navaneeth/Projects/openmpw/CFU-Playground/soc/build/digilent_arty.proj_template/gateware/digilent_arty.net
    [X] pcf:  MISSING
    [S] place:  place -> /home/navaneeth/Projects/openmpw/CFU-Playground/soc/build/digilent_arty.proj_template/gateware/digilent_arty.place
    [S] place_constraints:  place_constraints -> /home/navaneeth/Projects/openmpw/CFU-Playground/soc/build/digilent_arty.proj_template/gateware/digilent_arty.preplace
    [S] route:  route -> /home/navaneeth/Projects/openmpw/CFU-Playground/soc/build/digilent_arty.proj_template/gateware/digilent_arty.route
    [S] sdc:  synth -> /home/navaneeth/Projects/openmpw/CFU-Playground/soc/build/digilent_arty.proj_template/gateware/digilent_arty.sdc
[...]

PR #810, opened 2024-01-10
Pin Yosys back to 0.29 as workaround for Nexus flow (Yosys #4081).

tcal-x:
Workaround for https://github.com/YosysHQ/yosys/issues/4081.

This issue is specific to the Lattice Nexus flow.

With current Yosys, the synthesized netlist from `synth_nexus` contains an explicit 1'b0 driving the low order CIN of a chain of CCU2 (carry chain) primitives.  This causes a `nextpnr-nexus` error.    `nextpnr-nexus` expects this input to be undriven which I assume implies an implicit 1'b0 input.

### DIFF
--- a/conf/environment-symbiflow.yml
+++ b/conf/environment-symbiflow.yml
@@ -13,7 +13,8 @@ dependencies:
   - litex-hub::nextpnr-ecp5
   - litex-hub::nextpnr-ice40
 # Temporarily pin Yosys until symbiflow-yosys-plugins catches up
-  - litex-hub::yosys=0.33_11_g31ee566ec=20230724_080446_py37
+#   AND https://github.com/YosysHQ/yosys/issues/4081 is resolved
+  - litex-hub::yosys=0.29_35_g57c9eb70f=20230524_010725_py37
   - litex-hub::iceprog
   - litex-hub::prjxray-tools
   - litex-hub::prjxray-db
--- a/conf/environment.yml
+++ b/conf/environment.yml
@@ -13,7 +13,8 @@ dependencies:
   - litex-hub::nextpnr-ice40
   - litex-hub::iceprog
 # Temporarily pin Yosys until symbiflow-yosys-plugins catches up
-  - litex-hub::yosys=0.33_11_g31ee566ec=20230724_080446_py37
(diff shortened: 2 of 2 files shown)


tcal-x:
Hi @PiotrZierhoffer!   I still seem to be seeing intermittent certificate issues with dl.antmicro.com.  I could probably get it to green if I could selectively rerun the failing test(s), but unfortunately I don't have the permission to do so.

proppy:
> This causes a nextpnr-nexus error. nextpnr-nexus expects this input to be undriven which I assume implies an implicit 1'b0 input.

I'm curious why this error is not showing up in the CI, are we missing some tests?
https://github.com/google/CFU-Playground/actions/runs/7270368055

tcal-x:
I'm not sure where that link is from, but test-projects.yml is just one of the CI tests; it only runs Renode simulation and doesn't run synth / place / route for any platforms, so it doesn't encounter the issue here.   Only the oxide-ci test in oxide.yml and other tests building bitstreams for the Nexus FPGA will encounter the issue.

proppy:
Looks like https://github.com/google/CFU-Playground/actions/workflows/oxide.yml?query=branch%3Amain++ was never ran on main, do we have a existing run of this job failing? (or can you share instruction to reproduce this locally? is that similar to #808?)

tcal-x:
Here is a run from the semi-automatic Renode update, from within the last 24hrs -- it is failing on the issue that's fixed by this PR (`CCU2`):

https://github.com/google/CFU-Playground/actions/runs/8368656346/job/22913100078

image

tcal-x:
But I agree, the CI runs could use a review and reorg, to make it easier to find the ongoing failures.

tcal-x:
> But I agree, the CI runs could use a review and reorg, to make it easier to find the ongoing failures.

I wasn't planning on doing that in this PR though, @proppy.

proppy:
Makes sense, I just wanted a way to check that it's indeed fixing the issue:
- before 🔴  https://github.com/google/CFU-Playground/actions/runs/8368656346/job/22913100078
- after 🍏  https://github.com/google/CFU-Playground/actions/runs/7493242461/job/20398519775?pr=810

proppy (approved):

PR #817, opened 2024-04-17
Fix usage of read command.

offnaria:
Hello.
The `read` command returns nothing, so the expression in the following `apt` commands are executed whatever you write to the prompt.
I fixed it, and the `apt` commands are executed when you type `y` or nothing.
If the conventional behavior is expected, please close this PR.
Thank you.


offnaria:
I noticed this is the same as #769.

